PR #4, opened 2018-10-05
Prepare for tenancy - add tenants table and tenant_id to all current tables

gtanzillo:


agrare:
Should we create at least a simple tenant model also?  Maybe something we can lookup from a userid string and then set the tenant_id in refresh.

gtanzillo:
> Should we create at least a simple tenant model also? Maybe something we can lookup from a userid string and then set the tenant_id in refresh.

Makes sense, I'll add that

db/migrate/20181005182645_add_tenant_and_tenant_id_to_all_tables.rb
```diff
@@ -0,0 +1,20 @@
+class AddTenantAndTenantIdToAllTables < ActiveRecord::Migration[5.0]
+  create_table "tenants", id: :bigserial, force: :cascade do |t|
```
agrare: Shouldn't this be in the `change` method?

gtanzillo: Doh, you're right. I threw it in last minute. I'll take care of that and also add the model.

gtanzillo:
@agrare This on should be ready to go. Drew and I worked out the schema for the `tenants` table.

PR #5, opened 2018-10-08
Fix issue with running db:migrate from main app

agrare:
Add db/migrate paths in the engine to the app's path list

### DIFF
--- a/lib/topological_inventory/engine.rb
+++ b/lib/topological_inventory/engine.rb
@@ -1,5 +1,13 @@
 module TopologicalInventory
   class Engine < ::Rails::Engine
     isolate_namespace TopologicalInventory
+
+    initializer :append_migrations do |app|
+      config.paths["db/migrate"].expanded.each do |expanded_path|
+        app.config.paths["db/migrate"] << expanded_path
+      end
+      ActiveRecord::Migrator.migrations_paths = app.config.paths["db/migrate"]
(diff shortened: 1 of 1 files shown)


agrare:
I had that at first but it wasn't working for some reason, might be because I had a symlink to the engine from the main app.  Let me try pointing to this branch in the main Gemfile and see if that's why

agrare:
Hm yeah that was it, for some reason when using a symlink that evaluates to false and it doesn't add the paths.

agrare:
Duh it is because it is doing a regex match for some reason and now `root.to_s` is a substring of `app.root.to_s`.
How about doing a strict equality check instead of a regex match?

carbonin:
`is app.root.to_s` just the root directory or the full path?

agrare:
The full paths:
```
(byebug) app.root.to_s
"/home/agrare/src/topological_inventory/topological_inventory-api"
(byebug) root.to_s
"/home/agrare/src/topological_inventory/topological_inventory"
```

carbonin:
I'm good with it either way. I don't see how a change from `match` to `==` could really break anything. They're either the same location or not, I don't really understand a situation where we would benefit from the regex version, right?

agrare:
Yeah I'm not sure why manageiq-schema used match in the first place, it is just part of the `Initial changes to get things working.` commit with no other explanation.

PR #10, opened 2018-10-11
Add ContainerNodes and source_created_at/source_deleted_at

agrare:
This adds the ContainerNodes model with a reference to ContainerGroups, and also adds source_created_at and source_deleted_at attributes to all container/service models.

There was already a deleted_at which is what we had in MIQ but since it is coming from the source data I thought it would be better to keep it consistent with source_created_at (MIQ calls this ems_created_on).

### DIFF
--- a/db/migrate/20181011164429_add_container_nodes.rb
+++ b/db/migrate/20181011164429_add_container_nodes.rb
@@ -1,0 +1,18 @@
+class AddContainerNodes < ActiveRecord::Migration[5.1]
+  def change
+    create_table "container_nodes", :id => :bigserial do |t|
+      t.references "source", :type => :bigint, :index => true
+      t.string     "source_ref"
+      t.string     "resource_version"
+      t.string     "name"
+      t.integer    "cpus"
+      t.bigint     "memory"
+      t.references "tenant"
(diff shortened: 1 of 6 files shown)


db/migrate/20181011164429_add_container_nodes.rb
```diff
@@ -0,0 +1,18 @@
+class AddContainerNodes < ActiveRecord::Migration[5.1]
+  def change
+    create_table "container_nodes", :id => :bigserial do |t|
+      t.references "source", :type => :bigint, :index => true
+      t.string     "source_ref"
+      t.string     "resource_version"
+      t.string     "name"
+      t.integer    "cpus"
+      t.bigint     "memory"
```
agrare: I wasn't sure if we wanted to go full-blown ComputerSystem + Hardware with this or just keep it simple, thoughts?

agrare:
@gtanzillo @Ladas @bdunne @carbonin please review

db/migrate/20181011164429_add_container_nodes.rb
Ladas: yeah, lets keep it simple

PR #14, opened 2018-10-15
[WIP] Add an ArchivedConcern and scopes

agrare:
Add active and archived scopes for managed entities

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -12,13 +12,13 @@ class Source < ApplicationRecord
 
 
   # Container Inventory Objects
-  has_many :container_groups,    :dependent => :destroy
-  has_many :container_templates, :dependent => :destroy
-  has_many :container_projects,  :dependent => :destroy
-  has_many :container_nodes,     :dependent => :destroy
+  has_many :container_groups,    -> { active }, :dependent => :destroy
+  has_many :container_templates, -> { active }, :dependent => :destroy
+  has_many :container_projects,  -> { active }, :dependent => :destroy
+  has_many :container_nodes,     -> { active }, :dependent => :destroy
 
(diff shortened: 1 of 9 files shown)


agrare:
So this looks at `:source_deleted_at` but @Ladas brought up a good point that if we miss the notice and delete an entity from a full refresh and set this to `Time.now` it won't be the time it was deleted from the source.  The suggestion was to also have an `:archived_at` which is when it was archived in our DB allowing `:source_deleted_at` to be nil if we missed the notice.

app/models/source.rb
```diff
@@ -12,13 +12,13 @@ def default_endpoint
 
 
   # Container Inventory Objects
-  has_many :container_groups,    :dependent => :destroy
-  has_many :container_templates, :dependent => :destroy
-  has_many :container_projects,  :dependent => :destroy
-  has_many :container_nodes,     :dependent => :destroy
+  has_many :container_groups,    -> { active }, :dependent => :destroy
```
Ladas: so, if everything is targeted, this might not be needed?
So instead, this would be an API filter? Or I am trying to think about how will we query all these.

agrare: Without this we always pick up records that were already deleted when doing the save:

e.g.: `D, [2018-10-15T13:51:58.266408 #19477] DEBUG -- : Processing InventoryCollection:<ContainerGroup>, created=0, updated=102, deleted=46...Complete`
These 46 were already deleted

Ladas: that shouldn't be happening if the :targeted is true

Ladas:
ok, we can close this I've finished this in https://github.com/ManageIQ/topological_inventory-core/pull/19

PR #20, opened 2018-10-18
Add foreign key constraint to all foreign keys

Ladas:
Add foreign key constraint to all foreign keys. Having referential integrity in the DB is crucial for having the data in a good shape.


Ladas:
@agrare we should write some spec checking all foreign keys have the constraint, not sure how to do that yet

carbonin:
@Ladas I'm not opposed to this, but it's quite a difference from what we were doing in ManageIQ. Why the change? Is this just to get the cascading delete on the database side rather than using AR associations for that?

Ladas:
@carbonin Yes :-) We'll not use rails for DB operations :-) A lot of things will be different than in ManageIQ I suppose. :-)

carbonin:
Yeah, that's fine, I agree with you, but it's always good to include the "why" behind changes you're making.

db/migrate/20181018085055_add_foreign_keys_to_all_models.rb
```diff
@@ -0,0 +1,37 @@
+class AddForeignKeysToAllModels < ActiveRecord::Migration[5.1]
+  def change
+    # On cascade delete
```
carbonin: "On delete cascade" I think makes more sense.

db/migrate/20181018085055_add_foreign_keys_to_all_models.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    add_foreign_key :container_groups, :container_nodes, on_delete: :cascade
+
+    add_foreign_key :sources, :tenants, on_delete: :cascade
+
+    add_foreign_key :authentications, :tenants, on_delete: :cascade
+
+    add_foreign_key :service_parameters_sets, :service_offerings, on_delete: :cascade
+
+    # On cascade nullify
```
carbonin: "On delete nullify" I think makes more sense.

Ladas: right

app/models/service_parameters_set.rb
```diff
@@ -3,5 +3,5 @@ class ServiceParametersSet < ApplicationRecord
   belongs_to :source
   belongs_to :service_offering
 
-  has_many   :service_instances, :dependent => :nullify
```
Ladas: @agrare I've changed this to on delete cascade, I think a plan can't live without offering

app/models/container_node.rb
```diff
@@ -1,5 +1,5 @@
 class ContainerNode < ApplicationRecord
   belongs_to :tenant
   belongs_to :source
-  has_many   :container_groups, :dependent => :nullify
```
Ladas: @agrare changed this to delete, I think pod can't live without a node to run on?

Ladas:
@carbonin ok, added more description to the PR. :-)

app/models/service_parameters_set.rb
agrare: Possibly?  They are a separate top-level collection though so deletion should probably be handled by refresh.

PR #22, opened 2018-10-18
Add timestamps allowing atomic checks for parallel saving

Ladas:
If saving the same data in parallel, only the latest version
will be saved, even if saving partial row.


agrare:
> db/migrate/20181029172858_add_timestamps_for_parallel_saving.rb

Coming to you from the future?!?

Ladas:
@agrare I suppose that is the estimated time of merge :-)

PR #24, opened 2018-10-19
Spec tests for TopologicalInventory::Persister

agrare:
Add Rakefile tasks for spec:setup, db:create, db:drop, and db:migrate, update .travis.yml, and add a bin/setup so that we can run spec tests against the test db.

Test persister with mock inventory stored in json files to test saving simple inventory as well as saving associations between inventory collections.


agrare:
@bdunne @carbonin can you take a look?

spec/tools/test_inventory_generator
```diff
@@ -0,0 +1,75 @@
...
+            :reference                 => {:name => container_node.name},
+            :ref                       => :by_name,
+          )
+        )
+      ]
+    ),
+  ]
+)
+
+File.write(File.join(test_inventory_dir, "relationships_inventory.json"), JSON.dump(relationships_inventory.to_hash))
```
bdunne: Can all of this live in a module with something like `TestInventoryGenerator.run if __FILE__ == $PROGRAM_NAME` at the end?  Then if the file is required, it won't unintentionally do anything.

bin/setup
```diff
@@ -0,0 +1,28 @@
...
+def system!(*args)
+  system(*args) || abort("\n== Command #{args} failed ==")
+end
+
+chdir APP_ROOT do
+  # This script is a starting point to setup your application.
+  # Add necessary setup steps to this file.
+
+  puts '== Installing dependencies =='
+  system! 'gem install bundler --conservative'
```
bdunne: I'm wondering if we're going to need as much here as we did in ManageIQ ¯\\\_(ツ)_/¯

.travis.yml
```diff
@@ -10,3 +10,6 @@ addons:
 before_install:
 - gem install bundler -v 1.16.4
 - psql -c "CREATE USER root SUPERUSER PASSWORD 'smartvm';" -U postgres
+before_script:
+- cp config/database.dev.yml config/database.yml
```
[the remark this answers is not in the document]
Fryguy: This seems ok for Travis

bin/setup
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env ruby
+require 'pathname'
+require 'fileutils'
+include FileUtils
+
+# path to your application root.
+APP_ROOT = Pathname.new File.expand_path('../../', __FILE__)
```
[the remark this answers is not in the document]
Fryguy: If you need a pathname, even more preferable might be

```ruby
Pathname.new(__dir_).join("..").expand_path
```

Rakefile
```diff
@@ -20,6 +20,44 @@ require 'bundler/gem_tasks'
 
 require "rspec/core/rake_task"
 
+require "active_record"
+
+root = File.expand_path(__dir__)
```
Fryguy: Perhaps prefer a Pathname here to make the later usage of it cleaner?

.travis.yml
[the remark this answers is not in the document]
Fryguy: That being said, we did the following in service_portal, which also allowed for different user/pass for local DBs: 

https://github.com/ManageIQ/service_portal-api/blob/45541d9ac39968925c5800573441cf97b9689419/config/database.yml#L5-L6
https://github.com/ManageIQ/service_portal-api/blob/master/.travis.yml#L13-L14

Curious on your thoughts.  In any case, I'd like the .travis.yml files to be as close to the same as possible to make it easier on scripting changes

Rakefile
agrare: :+1: yeah that's a lot better

Rakefile
```diff
@@ -20,6 +20,44 @@ require 'bundler/gem_tasks'
...
+ActiveRecord::Tasks::DatabaseTasks.env = ENV["ENV"] || "development"
+ActiveRecord::Tasks::DatabaseTasks.database_configuration = YAML.load(File.read(File.join(root, 'config/database.yml')))
```
Fryguy: Prefer `YAML.load_file(filename)`

bin/setup
agrare: Hopefully not :crossed_fingers:

Rakefile
agrare: Yeah already changed to that in ff225839d851921ec9e0585cb9ce3995e3c5fabe

spec/tools/test_inventory_generator
agrare: Yeah sure thing

Rakefile
```diff
@@ -20,6 +20,44 @@ require 'bundler/gem_tasks'
...
+# Spec related rake tasks
 RSpec::Core::RakeTask.new(:spec)
 
+namespace :spec do
+  task :initialize do
+    ActiveRecord::Tasks::DatabaseTasks.env = "test"
+  end
+
+  desc "Setup the database for running tests"
+  task :setup => ["initialize", "db:create", "db:migrate"]
```
Fryguy: We technically don't need to db:create / db:migrate like ManageIQ, so we could possibly commit schema.rb and then it's just db:reset

.travis.yml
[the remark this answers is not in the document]
bdunne: Sounds like even more of a reason to standardize on `DATABASE_URL`

bin/setup
```diff
@@ -0,0 +1,28 @@
...
+
+def system!(*args)
+  system(*args) || abort("\n== Command #{args} failed ==")
```
Fryguy: Printing out the args like this will print it out like an Array...you probably want something like `Command #{args.join(" ")} failed`

.travis.yml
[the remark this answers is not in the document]
Fryguy: @syncrou Can you comment here why you made the change above?

lib/ar_helper.rb
```diff
@@ -7,7 +7,7 @@
 require "application_record"
 Dir[File.expand_path("app/models/**/*.rb", app_root)].each { |f| require f }
 
-app_env = ENV["TOPOLOGICAL_INVENTORY_ENV"] || "development"
+app_env = ENV["ENV"] || "development"
```
Fryguy: RAILS_ENV maybe? ... this particular env var now feels too general.

lib/topological_inventory/persister/logging.rb
```diff
@@ -0,0 +1,17 @@
+module TopologicalInventory
+  module Persister
+    class << self
+      attr_writer :logger
+    end
+
+    def self.logger
+      @logger ||= Logger.new(STDOUT, :level => Logger::ERROR)
```
Fryguy: Eventually we are going to need to do common logging...we should create a work task for that.

lib/ar_helper.rb
agrare: :+1: done

lib/topological_inventory/persister/logging.rb
bdunne: Just need a gem name.  `top_logic-logging`?

agrare:
> Maybe it should be in a module so it's invokable? That way if we load the file it doesn't start executing?

Yeah that works, something like `TopologicalInventory::Persister::ArHelper.load_environment!` when we start the worker?

bin/setup
agrare: :+1: I pulled this from the swagger generated `bin/setup` in [topological_inventory-api/bin/setup](https://github.com/ManageIQ/topological_inventory-api/blob/master/bin/setup#L10) I'll push a PR to fix it there as well.

agrare:
This was getting pretty big so I split out two other PRs for the database rake tasks (https://github.com/ManageIQ/topological_inventory-core/pull/25) and the ArHelper namespacing (https://github.com/ManageIQ/topological_inventory-core/pull/26)

spec/persister/worker_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      allow(ManageIQ::Messaging::Client).to receive(:open).and_return(client)
+      allow(client).to receive(:close)
+      allow(client).to receive(:subscribe_topic).and_yield(nil, nil, inventory)
+
+      described_class.new.run
+      source.reload
+    end
+
+    context "with a simple payload" do
+      let(:inventory) { JSON.load(File.read(File.expand_path(File.join("test_inventory", "simple_inventory.json"), __dir__))) }
```
bdunne: Is the `expand_path` necessary?

spec/persister/worker_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        )
+      end
+
+      it "sets the tenant to the source's tenant" do
+        expect(source.container_projects.first.tenant).to eq(source.tenant)
+      end
+    end
+
+    context "with some inter-object relationships" do
+      let(:inventory) { JSON.load(File.read(File.expand_path(File.join("test_inventory", "relationships_inventory.json"), __dir__))) }
```
bdunne: same here

spec/tools/test_inventory_generator
```diff
@@ -0,0 +1,83 @@
...
+
+require "json"
+
+require "topological_inventory/ingress_api/client"
+include TopologicalInventory::IngressApi::Client
+
+module TopologicalInventory
+  module TestInventoryGenerator
+    def self.run
+      test_inventory_dir = File.expand_path(File.join("..", "persister", "test_inventory"), __dir__)
```
bdunne: same here

PR #25, opened 2018-10-22
Database rake tasks for topological_inventory-core

agrare:
This adds a number of database rake tasks to assist in development in topological_inventory-core without the full rails env from topological_inventory-api


bin/setup
```diff
@@ -0,0 +1,28 @@
...
+  # This script is a starting point to setup your application.
+  # Add necessary setup steps to this file.
+
+  puts '== Installing dependencies =='
+  system! 'gem install bundler --conservative'
+  system('bundle check') || system!('bundle install')
+
+  puts "\n== Copying sample files =="
+  unless File.exist?('config/database.yml')
+    cp 'config/database.dev.yml', 'config/database.yml'
```
bdunne: Is this supposed to use `FileUtils`?

bdunne: nevermind, I missed the `include` above

db/schema.rb
```diff
@@ -0,0 +1,201 @@
...
+# you'll amass, the slower it'll run and the greater likelihood for issues).
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 20181018082155) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
+
+  create_table "authentications", force: :cascade do |t|
```
bdunne: Does this create the ids as `bigserial`?

agrare: Yes,
```
topological_inventory_development=# \d+ authentications
                                                         Table "public.authentications"
     Column     |       Type        | Collation | Nullable |                   Default                   | Storage  | Stats target | Description 
----------------+-------------------+-----------+----------+---------------------------------------------+----------+--------------+-------------
 id             | bigint            |           | not null | nextval('authentications_id_seq'::regclass) | plain    |              | 
```

Looks like in 5.1 `:bigserial` is the default primary key, and in schema_dumper they don't skip the `:id` if it is default: https://github.com/rails/rails/blob/v5.1.6/activerecord/lib/active_record/connection_adapters/abstract/schema_dumper.rb#L13-L14 and https://github.com/rails/rails/blob/v5.1.6/activerecord/lib/active_record/connection_adapters/postgresql/schema_dumper.rb#L19-L21

PR #28, opened 2018-10-23
Ensure that Endpoint#default is defaulted to false

bdunne:


bdunne:
> Hey @bdunne can you `rake db:migrate` and commit the updated `db/schema.rb` ?

:+1: Done

PR #32, opened 2018-10-24
Add a SourceType model for dynamic provider types

agrare:
Rather than have a string column for the source type and a pre-defined list of known types, add a table so that source types can be added dynamically along with the json schema for the create/update forms.  Then we can seed the builtin source types like openshift/aws/etc...

### DIFF
--- a/app/models/source.rb
+++ b/app/models/source.rb
@@ -1,5 +1,6 @@
 class Source < ApplicationRecord
   has_many :endpoints, :autosave => true
+  belongs_to :source_type
   belongs_to :tenant
 
   delegate :scheme, :scheme=, :host, :host=, :port, :port=, :path, :path=,
--- a/db/migrate/20181024163850_add_source_type.rb
+++ b/db/migrate/20181024163850_add_source_type.rb
@@ -8,6 +8,6 @@ class AddSourceType < ActiveRecord::Migration[5.1]
       t.index :name, :unique => true
     end
 
-    add_reference :sources, :source_type, :index => true
+    add_reference :sources, :source_type,  :index => true, :foreign_key => {:on_delete => :cascade}
   end
 end
--- a/db/migrate/20181024163850_add_source_type.rb
+++ b/db/migrate/20181024163850_add_source_type.rb
@@ -1,0 +1,13 @@
+class AddSourceType < ActiveRecord::Migration[5.1]
+  def change
+    create_table :source_types, :id => :bigserial do |t|
+      t.string :name
+      t.string :product_name
+      t.string :vendor
+      t.timestamps
+      t.index :name, :unique => true
+    end
+
+    add_reference :sources, :source_type, :index => true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -230,12 +230,23 @@ ActiveRecord::Schema.define(version: 20181102145252) do
     t.index ["source_id"], name: "index_source_regions_on_source_id"
   end
 
+  create_table "source_types", force: :cascade do |t|
+    t.string "name"
+    t.string "product_name"
+    t.string "vendor"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["name"], name: "index_source_types_on_name", unique: true
+  end
+
   create_table "sources", force: :cascade do |t|
     t.string "name"
     t.string "uid"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.bigint "tenant_id", null: false
+    t.bigint "source_type_id"
+    t.index ["source_type_id"], name: "index_sources_on_source_type_id"
   end
 
   create_table "subscriptions", force: :cascade do |t|
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -336,6 +336,7 @@ ActiveRecord::Schema.define(version: 20181102145252) do
   add_foreign_key "service_plans", "subscriptions", on_delete: :cascade
   add_foreign_key "service_plans", "tenants", on_delete: :cascade
   add_foreign_key "source_regions", "sources", on_delete: :cascade
+  add_foreign_key "sources", "source_types", on_delete: :cascade
   add_foreign_key "sources", "tenants", on_delete: :cascade
   add_foreign_key "subscriptions", "sources", on_delete: :cascade
 end
--- a/app/models/source_type.rb
+++ b/app/models/source_type.rb
@@ -1,0 +1,3 @@
+class SourceType < ApplicationRecord
+  has_many :sources
+end
--- a/spec/persister/worker_spec.rb
+++ b/spec/persister/worker_spec.rb
@@ -2,9 +2,17 @@ require "topological_inventory/persister/worker"
 
 describe TopologicalInventory::Persister::Worker do
   let(:tenant) { Tenant.find_or_create_by!(:name => "default") }
+  let(:source_type) { SourceType.find_or_create_by(:name => "openshift") }
   let(:client) { double(:client) }
-  let!(:source) { Source.find_or_create_by!(:name => "OCP", :uid => "9a874712-9a55-49ab-a46a-c823acc35503", :tenant => tenant) }
   let(:test_inventory_dir) { Pathname.new(__dir__).join("test_inventory") }
+  let!(:source) do
+    Source.find_or_create_by!(
+      :tenant      => tenant,
+      :source_type => source_type,
+      :name        => "OCP",
+      :uid         => "9a874712-9a55-49ab-a46a-c823acc35503",
+    )
+  end
 
   context "#run" do
     let(:messages) { [ManageIQ::Messaging::ReceivedMessage.new(nil, nil, inventory, nil)] }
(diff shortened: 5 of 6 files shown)


agrare:
Is a single json schema is enough to handle everything we need for a provider screen?  Thinking possible endpoints, authentications types, etc...

db/migrate/20181024163850_add_source_type.rb
```diff
@@ -0,0 +1,15 @@
+class AddSourceType < ActiveRecord::Migration[5.1]
+  def change
+    create_table :source_types, :id => :bigserial do |t|
+      t.string :name
+      t.string :display_name
+      t.string :vendor
+      t.jsonb :create_schema
+      t.jsonb :update_schema
```
bdunne: I'm trying to come up with a use case for this.  Maybe a required field that is only populated on create then read-only after that?

db/migrate/20181024163850_add_source_type.rb
```diff
@@ -0,0 +1,15 @@
+class AddSourceType < ActiveRecord::Migration[5.1]
+  def change
+    create_table :source_types, :id => :bigserial do |t|
```
bdunne: Should this be more generic?  Maybe json_schemas?  I'm thinking we may reuse this on Endpoints and Authentications, right?

db/migrate/20181024163850_add_source_type.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.string :name
+      t.string :display_name
```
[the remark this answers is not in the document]
agrare: I was thinking name would be the internal name like "openshift" or "vsphere" and the display name would be "Openshift Container Platform" or "VMware vSphere".  The goal being we could update the display name for e.g. rebrandings but keep the internal name the same, and the internal name is what the unique constraint is on.  Maybe `:type` would be a better column name?

db/migrate/20181024163850_add_source_type.rb
```diff
@@ -0,0 +1,15 @@
...
+      t.string :vendor
```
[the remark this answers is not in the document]
agrare: For Vendor I was thinking `Red Hat`, `Microsoft`, etc... so that we split the "company" from the "product".  This is a bit confusing in MIQ because we aren't consistent in what is the vendor vs what is the product: Microsoft::InfraManager => SCVMM by Microsoft and Azure::CloudManager => Azure by Microsoft, Redhat::InfraManager => RHV by Red Hat and Openshift::ContainerManager => Openshift by Red Hat

db/migrate/20181024163850_add_source_type.rb
bdunne: I wonder if we would be able to handle that with a conditional https://json-schema.org/draft-07/json-schema-validation.html#rfc.section.6.6

db/migrate/20181024163850_add_source_type.rb
agrare: That would kind of pollute the table if json_schemas.name could include "openshift" and "ssh" without some other type as well.  Maybe we need an authentication_types also?

db/migrate/20181024163850_add_source_type.rb
agrare: We can go with just `:schema` then if one field can handle it

bdunne: I haven't tested conditionals, but I'd prefer to go with one column until we have a hard requirement for the second.

bdunne:
Where would we gather the seed data from?

agrare:
@bdunne I was thinking `db/seeds.rb` in the core repo

bdunne:
Is it worth storing them in the database if they're also on the filesystem?

db/migrate/20181024163850_add_source_type.rb
carbonin: Agree with @agrare here, even if we don't need separate tables for what we're storing I think this is the better approach. I would rather have a few tables with the same kind of data if it avoids the kind of confusion we would need to deal with to solve a few special cases.

carbonin:
@bdunne it's not so easy to join across the filesystem :trollface: 

But in all seriousness what would the alternative be? Especially if this is something we expect to be created dynamically in the future?

db/migrate/20181024163850_add_source_type.rb
```diff
@@ -0,0 +1,14 @@
...
+      t.jsonb :schema
+      t.timestamps
+      t.index :name, :unique => true
+    end
+
+    add_reference :sources, :source_type, :index => true, :null => false
```
Ladas: ```add_foreign_key :sources, :source_type, on_delete: :cascade```

Not sure if we should do cascade delete, if somebody deletes a source type?

agrare: Yeah that comes down to if we want to allow sources without an associated type.  I'm leaning towards not because then we open the door for third parties to be lazy and not explicitly define the source type.

app/models/source.rb
```diff
@@ -1,4 +1,5 @@
 class Source < ApplicationRecord
+  belongs_to :source_type
   has_many :endpoints, :autosave => true
   belongs_to :tenant
```
Fryguy: aesthetically, keep the belongs_tos together

db/migrate/20181024163850_add_source_type.rb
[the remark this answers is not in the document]
Fryguy: `type` is for STI.

db/schema.rb
```diff
@@ -225,12 +225,24 @@
...
+    t.index ["name"], name: "index_source_types_on_name", unique: true
+  end
+
   create_table "sources", force: :cascade do |t|
     t.string "name"
     t.string "uid"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.bigint "tenant_id", null: false
+    t.bigint "source_type_id", null: false
```
Fryguy: If this is null false, you might need a data migration to fix existing records.

lib/topological_inventory/core/seeding.rb
```diff
@@ -0,0 +1,12 @@
+module TopologicalInventory
+  module Core
+    class Seeding
+      def self.seed!
+        SourceType.create_with(
+          :display_name => "Red Hat OpenShift Container Platform",
+          :vendor       => "Red Hat, Inc.",
+        ).find_or_create_by!(:name => "openshift")
```
Fryguy: I don't like having pluggable things in this repo.

db/migrate/20181024163850_add_source_type.rb
carbonin: > Is a single json schema is enough to handle everything we need for a provider screen? Thinking possible endpoints, authentications types, etc...

Reread this comment ... is the json schema in this table also going to include the spec to create the source type's authentication and endpoint? Or is that going to be separate?

If it's all rolled into one schema then I think what we have here is what @bdunne is asking for right?

lib/topological_inventory/core/seeding.rb
carbonin: Agree, but it would be good to get a demo schema in a comment or something to see what the intended use would be like.

db/migrate/20181024163850_add_source_type.rb
agrare: > Reread this comment ... is the json schema in this table also going to include the spec to create the source type's authentication and endpoint? Or is that going to be separate?

Yeah that is what I'm wondering, I'm not sure how the UI will use this.  If all in one schema works then great, otherwise we'll have to specify what endpoint/auth types a source_type needs here, then we'll have to go to an endpoint_type.schema.

@martinpovolny do you know if a single json spec for a provider create screen will work?

bdunne: I was expecting separate json schemas for sources and endpoints (and sources referring to endpoints).

lib/topological_inventory/core/seeding.rb
agrare: Okay i'll drop the seeding for now

db/migrate/20181024163850_add_source_type.rb
agrare: Okay we can handle endpoint type json schema in a follow-up pr

db/migrate/20181024163850_add_source_type.rb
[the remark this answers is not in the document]
bdunne: product_name?

db/migrate/20181024163850_add_source_type.rb
```diff
@@ -0,0 +1,14 @@
+class AddSourceType < ActiveRecord::Migration[5.1]
+  def change
+    create_table :source_types, :id => :bigserial do |t|
+      t.string :name
+      t.string :display_name
+      t.string :vendor
+      t.jsonb :schema
```
bdunne: Do we need this right now?

db/migrate/20181024163850_add_source_type.rb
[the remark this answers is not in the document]
agrare: :+1: works for me

db/migrate/20181024163850_add_source_type.rb
agrare: No I can remove it for now

db/schema.rb
agrare: Removed the non-null constraint until we figure out the seeding since we wouldn't have anything to set it to in a data migration

PR #36, opened 2018-10-26
Add a tasks model to track running operations

agrare:
Add a tasks model to keep track of long running async operations.

Dependent: https://github.com/ManageIQ/topological_inventory-api/pull/31

### DIFF
--- a/db/migrate/20181026182144_add_tasks_table.rb
+++ b/db/migrate/20181026182144_add_tasks_table.rb
@@ -4,6 +4,7 @@ class AddTasksTable < ActiveRecord::Migration[5.1]
       t.string :name
       t.string :status
       t.string :state
+      t.jsonb :context
       t.references :tenant, :index => true, :null => false
       t.datetime :completed_at
       t.timestamps
--- a/db/migrate/20181026182144_add_tasks_table.rb
+++ b/db/migrate/20181026182144_add_tasks_table.rb
@@ -1,0 +1,12 @@
+class AddTasksTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :tasks, :id => :bigserial do |t|
+      t.string :name
+      t.string :status
+      t.string :state
+      t.references :tenant, :index => true, :null => false
+      t.datetime :completed_at
+      t.timestamps
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -215,6 +215,7 @@ ActiveRecord::Schema.define(version: 20181026182144) do
     t.string "name"
     t.string "status"
     t.string "state"
+    t.jsonb "context"
     t.bigint "tenant_id", null: false
     t.datetime "completed_at"
     t.datetime "created_at", null: false
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20181023215716) do
+ActiveRecord::Schema.define(version: 20181026182144) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -209,6 +209,17 @@ ActiveRecord::Schema.define(version: 20181023215716) do
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.bigint "tenant_id", null: false
+  end
+
+  create_table "tasks", force: :cascade do |t|
+    t.string "name"
+    t.string "status"
+    t.string "state"
+    t.bigint "tenant_id", null: false
+    t.datetime "completed_at"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["tenant_id"], name: "index_tasks_on_tenant_id"
   end
 
   create_table "tenants", force: :cascade do |t|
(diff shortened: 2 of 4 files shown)


db/migrate/20181026182144_add_tasks_table.rb
```diff
@@ -0,0 +1,13 @@
+class AddTasksTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :tasks, :id => :bigserial do |t|
+      t.string :name
+      t.string :status
+      t.string :state
+      t.string :message
+      t.integer :progress
+      t.references :tenant, :index => true, :null => false
+      t.timestamps
```
bdunne: Should we also have a `completed_on` timestamp?

db/migrate/20181026182144_add_tasks_table.rb
```diff
@@ -0,0 +1,13 @@
+class AddTasksTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :tasks, :id => :bigserial do |t|
+      t.string :name
+      t.string :status
+      t.string :state
+      t.string :message
```
bdunne: I assume this will be custom messaging to go along with the progress?

db/migrate/20181026182144_add_tasks_table.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.integer :progress
```
bdunne: Will this be used in the initial implementation?

db/migrate/20181026182144_add_tasks_table.rb
```diff
@@ -0,0 +1,13 @@
+class AddTasksTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :tasks, :id => :bigserial do |t|
+      t.string :name
+      t.string :status
```
bdunne: Do we need this and message?

db/migrate/20181026182144_add_tasks_table.rb
```diff
@@ -0,0 +1,13 @@
...
+      t.string :state
```
Fryguy: What is the difference between status and state?

Fryguy:
We may want to consider an Array of statuses or something, so one can see the history of the task as it progresses.

db/schema.rb
```diff
@@ -178,28 +178,6 @@
...
-    t.jsonb "create_json_schema"
-    t.jsonb "update_json_schema"
-    t.datetime "archived_on"
-    t.index ["archived_on"], name: "index_service_parameters_sets_on_archived_on"
-    t.index ["service_offering_id"], name: "index_service_parameters_sets_on_service_offering_id"
-    t.index ["source_deleted_at"], name: "index_service_parameters_sets_on_source_deleted_at"
-    t.index ["source_id", "source_ref"], name: "index_service_parameters_sets_on_source_id_and_source_ref", unique: true
-    t.index ["source_id"], name: "index_service_parameters_sets_on_source_id"
-  end
-
```
Fryguy: I don't think this PR should be killing a table.

db/migrate/20181026182144_add_tasks_table.rb
agrare: Status would be e.g. ["success", "error", "warning"] and state could be ["queued", "running", "finished"]

db/migrate/20181026182144_add_tasks_table.rb
agrare: Yeah good idea

db/migrate/20181026182144_add_tasks_table.rb
agrare: Right this might be something that we would show as a summary on the UI

db/migrate/20181026182144_add_tasks_table.rb
agrare: Probably not, initially I expect we just mark the task completed when the service is deployed.

db/migrate/20181026182144_add_tasks_table.rb
agrare: We need status to know when the task is completed, we could do without message if we don't want to have any output initially.

db/schema.rb
agrare: Yeah this was fixed by https://github.com/ManageIQ/topological_inventory-core/pull/38 I just need to rebase

db/migrate/20181026182144_add_tasks_table.rb
```diff
@@ -0,0 +1,12 @@
+class AddTasksTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :tasks, :id => :bigserial do |t|
```
bdunne: We probably also need columns for `resource_type`, `resource_id`, `method` and `arguments` right?

agrare: `resource` so that we can link the task to what was created?  I don't want to limit it to that necessarily since there could be tasks that don't relate to any object or tasks which create multiple objects.  Maybe a jsonb context column ?

bdunne: No, how does the task know what method to call on which instance / class?

agrare: Oh I see, I wasn't thinking that we would have the task actually tell the operations worker what to do, it is primarily for tracking.  We might need an MiqQueue equivalent to do what you're talking about.

PR #37, opened 2018-10-26
Add basic tagging functionality

gtanzillo:
To Do:
- [X] Get spec test running in this repo - @agrare 
- [X] Add `acts_as_taggable` to the other models
- [X] Additional examples in specs
- [ ] ~Update Gemfile in `topological_inventory-api` repo?~

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,8 @@ source 'https://rubygems.org'
 # development dependencies will be added by default to the :development group.
 gemspec
 
+gem 'acts-as-taggable-on', '~> 6.0'
+
 # Declare any dependencies that are still in development here instead of in
 # your gemspec. These might include edge Rails or gems from your path or
 # Git. Remember to move these dependencies to your gemspec before releasing
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,4 +1,6 @@
 class ApplicationRecord < ActiveRecord::Base
+  require 'acts-as-taggable-on'
+
   self.abstract_class = true
 
   def as_json(options = nil)
--- a/spec/tagging/tagging_spec.rb
+++ b/spec/tagging/tagging_spec.rb
@@ -2,11 +2,6 @@ describe "ActsAsTaggableOn" do
   let(:tenant) { Tenant.find_or_create_by!(:name => "default") }
   let(:source) { Source.find_or_create_by!(:name => "OCP", :uid => "9a874712-9a55-49ab-a46a-c823acc35503", :tenant => tenant) }
 
-  after do
-    source.tag_list = []
-    source.save!
-  end
-
   context "common tagging operations" do
     it "is taggable" do
       expect(source.is_taggable?).to be_truthy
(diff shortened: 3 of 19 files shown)


gtanzillo:
Nice @agrare, that's easy enough. I'll add it to `ApplicationRecord`. I would expect every model to be taggable. Thanks for looking at it!

Gemfile
```diff
@@ -10,6 +10,7 @@ gemspec
 # Git. Remember to move these dependencies to your gemspec before releasing
 # your gem to rubygems.org.
 
+gem 'acts-as-taggable-on', '~> 6.0'
```
Fryguy: Please put this with the other gem list (the normal non-git ones).

Fryguy: Oh wait...I haven't updated the repo to enforce that yet :)

spec/tagging/tagging_spec.rb
```diff
@@ -0,0 +1,46 @@
+describe "ActsAsTaggableOn" do
+  let(:tenant) { Tenant.find_or_create_by!(:name => "default") }
+  let(:source) { Source.find_or_create_by!(:name => "OCP", :uid => "9a874712-9a55-49ab-a46a-c823acc35503", :tenant => tenant) }
+
+  after do
+    source.tag_list = []
+    source.save!
```
Fryguy: This shouldn't be necessary because the record should be destroyed.

spec/tagging/tagging_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      expect(source.tag_list).to eq(["totally", "freaking", "awesome"])
+
+      source.tag_list = []
+      source.save!
+
+      expect(source.tag_list).to eq([])
+    end
+  end
+end
```
Fryguy: Leave a CR after.

app/models/application_record.rb
```diff
@@ -1,4 +1,6 @@
 class ApplicationRecord < ActiveRecord::Base
+  require "acts_as_taggable_on"
```
Fryguy: curious why this is necessary since the Gemfile is requiring it

gtanzillo: For some reason it was not getting loaded. Without that line the tests fails like this
```
An error occurred while loading spec_helper.
Failure/Error: acts_as_taggable

NameError:
  undefined local variable or method `acts_as_taggable' for Source (call 'Source.connection' to establish a connection):Class
# /Users/gtanzill/.rvm/gems/ruby-2.3.1@vmdb_5_7/gems/activerecord-5.1.6/lib/active_record/dynamic_matchers.rb:22:in `method_missing'
# ./app/models/source.rb:8:in `<class:Source>'
# ./app/models/source.rb:1:in `<top (required)>'
# /Users/gtanzill/.rvm/gems/ruby-2.3.1@vmdb_5_7/gems/activesupport-5.1.6/lib/active_support/dependencies.rb:292:in `require'
# /Users/gtanzill/.rvm/gems/ruby-2.3.1@vmdb_5_7/gems/activesupport-5.1.6/lib/active_support/dependencies.rb:292:in `block in require'
# /Users/gtanzill/.rvm/gems/ruby-2.3.1@vmdb_5_7/gems/activesupport-5.1.6/lib/active_support/dependencies.rb:258:in `load_dependency'
# /Users/gtanzill/.rvm/gems/ruby-2.3.1@vmdb_5_7/gems/activesupport-5.1.6/lib/active_support/dependencies.rb:292:in `require'
# ./lib/topological_inventory/persister/ar_helper.rb:12:in `block in load_environment!'
# ./lib/topological_inventory/persister/ar_helper.rb:12:in `each'
# ./lib/topological_inventory/persister/ar_helper.rb:12:in `load_environment!'
# ./spec/spec_helper.rb:24:in `<top (required)>'

Randomized with seed 49817
```

PR #39, opened 2018-11-01
[WIP] Add ServicePlan#order and relevant external api changes

eclarizio:
This PR basically moves over code from https://github.com/ManageIQ/service_portal-api/blob/master/app/models/provider.rb#L95 to this repo onto the `ServicePlan` model.

A couple things I think I need some guidance on, since I mostly just moved over the code and refactored:

`service_plan#order` currently is returning `<some_parsed_external_data>['metadata']['selfLink']`. I think the original discussion was centered around wanting to return a `task_id` even if right now we're doing that task inline. So, this line will likely need to change but I'm not sure how. I don't think I know enough about what an actual response from the external api looks like or how we plan on delegating the work to a task.

Some of the variables in the new `ExternalApiBroker` (which we can discuss about the name, I just threw something together) are not correct or I guessed at. For example, the `verify_ssl` option is an attribute on the `Provider` class back in the service portal API, but I don't know if we have an equivalent here, so I simply hardcoded it to `true`. Same thing goes for the `token` attribute, as well as the `url` attribute, which I've just set as a placeholder here in `base_url`. Are these available as things the `ServicePlan` would know about somehow through a related object that I can pass in, or something else?

@agrare @gtanzillo @bdunne Can you guys review and advise on my questions above? Thanks!


app/models/service_plan.rb
```diff
@@ -1,4 +1,5 @@
 require "archived_concern"
+require_relative "../../lib/infrastructure/external_api_broker"
```
agrare: `./lib` is on the LOAD_PATH so you shouldn't need the require_relative

lib/infrastructure/external_api_broker.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def order_service_plan(payload)
+    make_request(method: :post, url: order_service_plan_url, payload: payload)
+  end
+
+  private
+
+  def base_url
+    "https://localhost"
```
agrare: This you can get from the source.default_endpoint

agrare: Can we add `apis/servicecatalog.k8s.io/v1beta1/` into the base_url?

lib/infrastructure/external_api_broker.rb
```diff
@@ -0,0 +1,63 @@
...
+  def plan_url
+    URI.join(base_url, "apis/servicecatalog.k8s.io/v1beta1/clusterserviceplans").to_s
+  end
+
+  def make_request(method:, url:, headers: generic_headers, payload: nil)
+    request_options = {
+      :method     => method,
+      :url        => url,
+      :headers    => headers,
+      :verify_ssl => true
```
agrare: We should add this to the default endpoint also

lib/infrastructure/external_api_broker.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    request_options.merge(:payload => payload) if payload.present?
+
+    RestClient::Request.new(request_options).execute
+  end
+
+  def generic_headers
+    token = "???"
+    {
+      "Authorization" => "Bearer #{token}", # Token?
```
agrare: You can get this from `source.default_endpoint.authentications.first.password`

lib/infrastructure/external_api_broker.rb
```diff
@@ -0,0 +1,63 @@
+require "rest_client"
+
+class ExternalApiBroker
```
agrare: Maybe `ServiceCatalogClient` ?  I don't care too much about the name though since its going to change soon anyway

agrare:
> @agrare I tried to separate the external API stuff too to make it a bit easier to change to kubeclient instead of RestClient. Hope that helps a bit at least smile

Appreciate it!

agrare:
@eclarizio so the task stuff got merged so you can create a task and return that ID.  There is a jsonb context there where you can put the id of the created object.  When we see what the return payload looks like we can figure that part out, for now lets just create the task record and return the ID

app/models/service_plan.rb
```diff
@@ -10,4 +11,44 @@ class ServicePlan < ApplicationRecord
   has_many   :service_instances
 
   acts_as_taggable
+
+  def order(catalog_id, additional_parameters)
```
bdunne: I feel like all of this should be in a module in the API repo since it's all temporary and nothing else here will ever use it.  It will be much more clear that we should delete all of it in the same PR that does it the newer way.

app/models/service_plan.rb
bdunne: It's probably not auto-loading because the constants are not in the `Infrastructure` namespace

lib/infrastructure/external_api_broker.rb
bdunne: We should probably go with something similar to `security_protocol` that we have on ManageIQ

agrare: Yeah that's what I was thinking too, also a custom certificate_authority.  I'll put a PR in to add both of those to Endpoints

app/models/service_plan.rb
eclarizio: Yeah we can talk about the directory naming structure, this was just something I've used in the past, and it happens to avoid something like `lib/services` since we have the name "service" in a bunch of actual objects, haha!

app/models/service_plan.rb
```diff
@@ -10,4 +11,45 @@ class ServicePlan < ApplicationRecord
...
+    response = service_catalog_client.order_service_plan(payload)
+
+    parsed_data = JSON.parse(response.body)
+    task = Task.create!(:tenant => tenant, :context => parsed_data['metadata']['selfLink'])
+    task.id
+  end
+
+  private
+
+  def external_payload(catalog_id, parameters)
```
agrare: This would be better in the service_catalog_client module IMO since it is very kube specific, you can just pass in the params to `#order_service_plan`

app/models/service_plan.rb
```diff
@@ -10,4 +11,45 @@ class ServicePlan < ApplicationRecord
   has_many   :service_instances
 
   acts_as_taggable
+
+  def order(catalog_id, additional_parameters)
+    payload = external_payload(catalog_id, additional_parameters)
+
+    response = service_catalog_client.order_service_plan(payload)
+
+    parsed_data = JSON.parse(response.body)
```
agrare: Could you parse the response.body and return the hash?  That would simplify this a good bit.

app/models/service_plan.rb
```diff
@@ -10,4 +11,45 @@ class ServicePlan < ApplicationRecord
...
+    task = Task.create!(:tenant => tenant, :context => parsed_data['metadata']['selfLink'])
```
agrare: How about:
```
:context => {:service_instance => {:source_id => ext_management_system.id, :source_ref => parsed_data['metadata']['selfLink']}
```

This will provide the caller enough information to find the new object.

app/models/service_plan.rb
agrare: I agree, if we move the rest of the methods to the service_catalog_client module I think that will accomplish this.  We can just add a comment over the `service_catalog_client.order_service_plan(payload)` call stating that this is temporary.

lib/infrastructure/external_api_broker.rb
agrare: Hey @eclarizio verify_ssl and certificate_authority were added to the Endpoints model in https://github.com/ManageIQ/topological_inventory-core/pull/42

eclarizio: Thanks @agrare I'll change this too, then.

lib/infrastructure/service_catalog_client.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def generic_headers
+    {
+      "Authorization" => "Bearer #{@source.default_endpoint.authentications.first.password}",
+      "Content-Type"  => "application/json",
+      "Accept"        => "application/json"
+    }
+  end
+
+  def get_object_name(object_id, url)
```
agrare: Do we have the plan/catalog names in our DB?  Seems like we should be able to save an API call to fetch something by id that we already have.

lib/infrastructure/service_catalog_client.rb
```diff
@@ -0,0 +1,86 @@
...
+    response = make_request(method: :get, url: url)
+
+    parsed_data = JSON.parse(response.body)
+    entry = parsed_data['items'].each.detect { |item| object_id == item['metadata']['name'] }
+    entry ? entry['spec']['externalName'] : nil
+  end
+
+  def build_service_plan_payload(plan_id, catalog_id, additional_parameters)
```
agrare: Hm you're passing in the database id (https://github.com/ManageIQ/topological_inventory-core/pull/39/files#diff-969bf8017ce44a90acb1589da617ac3eR18) and using it to request the name from kubernetes?  I don't think this will work.

app/models/service_plan.rb
```diff
@@ -12,4 +13,25 @@ class ServicePlan < ApplicationRecord
   has_many   :service_instances
 
   acts_as_taggable
+
+  def order(additional_parameters)
+    parsed_response = service_catalog_client.order_service_plan(name, service_offering.name, additional_parameters)
```
agrare: Now we're talking :)

lib/topological_inventory/core/service_catalog_client.rb
```diff
@@ -0,0 +1,59 @@
...
+module TopologicalInventory
+  module Core
+    class ServiceCatalogClient
+      def initialize(source)
+        @source = source
+      end
+
+      # TODO? Move this to a module in the API repo
+      def order_service_plan(plan_name, service_offering_name, additional_parameters)
+        payload = ServicePlanClient.new.build_payload(
```
agrare: I don't think we need a new client type for this but not a big deal I'm good with it for now

spec/models/service_plan_spec.rb
```diff
@@ -0,0 +1,44 @@
+describe ServicePlan do
+  describe "#order" do
+    let(:service_plan) do
+      described_class.new(:id               => 456,
```
agrare: Shouldn't need to specify an ID here, it'll be set automatically

agrare: Oh yeah, it'll be set automatically when you call `.create!` :) prefer that over `.new`

spec/models/service_plan_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+                          :source           => source,
+                          :tenant           => tenant,
+                          :name             => "plan_name",
+                          :service_offering => service_offering)
+    end
+    let(:source) { Source.new(:id => 654) }
```
agrare: Same here can just be `Source.create!(:tenant => tenant)`

spec/models/service_plan_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    let(:service_offering) { ServiceOffering.new(:name => "service_offering") }
```
agrare: Same

spec/models/service_plan_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    let(:tenant) { Tenant.new }
```
agrare: And this you can do `Tenant.find_or_create_by!(:name => "default")`

lib/topological_inventory/core/service_catalog_client.rb
eclarizio: You're probably right now that the payload doesn't need to actually make the external calls. When I shifted it out it was going to have to make them and was going to be a much bigger pain to write a spec for than I thought it should've been. Moving it out alleviated that issue but yeah after we figured out it doesn't even need to make the external calls it could probably be moved back in if we want.

lib/topological_inventory/core/service_catalog_client.rb
```diff
@@ -0,0 +1,59 @@
...
+          plan_name, service_offering_name, catalog_parameters(additional_parameters)
+        )
+        response = make_request(method: :post, url: order_service_plan_url, payload: payload)
+        JSON.parse(response.body)
+      end
+
+      private
+
+      def base_url
+        URI.join(@source.default_endpoint.host, "apis/servicecatalog.k8s.io/v1beta1/").to_s
```
bdunne: What about the rest of the URI components?

lib/topological_inventory/core/service_catalog_client.rb
```diff
@@ -0,0 +1,59 @@
...
+      def make_request(method:, url:, headers: generic_headers, payload: nil)
+        request_options = {
+          :method     => method,
+          :url        => url,
+          :headers    => headers,
+          :verify_ssl => @source.default_endpoint.verify_ssl,
+          :payload    => payload
+        }
+
+        request_options[:payload] = payload if payload.present?
```
bdunne: Should `payload` be removes from the hash above then?

spec/models/service_plan_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    it "orders the service plan" do
+      expect(service_catalog_client).to receive(:order_service_plan).with("plan_name", "service_offering", parameters)
+      service_plan.order(parameters)
+    end
+
+    it "creates a task with context information" do
+      service_plan.order(parameters)
+      expect(Task.last.context).to eq(
+        {:service_instance => {:source_id => source.id, :source_ref => 321}}.with_indifferent_access
```
bdunne: Why the `with_indifferent_access`?

spec/topological_inventory/core/service_catalog_client_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        before do
+          allow(ServicePlanClient).to receive(:new).and_return(service_plan_client)
+          allow(service_plan_client).to receive(:build_payload).with(
+            "plan_name", "service_offering_name", {"foo" => "bar", "baz" => "qux"}
+          ).and_return("payload")
+
+          allow(RestClient::Request).to receive(:new).with(post_request_options).and_return(dummy_rest_client)
+          allow(dummy_rest_client).to receive(:execute).and_return(double(:body => dummy_response.to_json))
+          allow(endpoint).to receive(:authentications).and_return([auth])
```
bdunne: I don't think we should be stubbing this

spec/models/service_plan_spec.rb
eclarizio: Was a mistake, it was failing without it so instead of changing the keys to strings I put that there to test if it would work and then forgot to change the keys when it did.

lib/topological_inventory/core/service_catalog_client.rb
eclarizio: Ah this logic needs to be re-written now that we don't have the `get` requests in this. Will fix.

lib/topological_inventory/core/service_catalog_client.rb
eclarizio: What do you mean? Or do you mean I should combine `base_url` and `order_service_plan_url` since nothing else uses `base_url`?

bdunne: `Endpoint.host` would be something like `"example.com"`.  This doesn't take into account the rest of the URL components that we store on the `Endpoint` for example:
- The OpenShift / Kubernetes api usually is not running on port 80/443
- HTTP vs HTTPS
- Any deviations to the base path

agrare:
Okay so when I run this with a randomly selected service plan it successfully posts to the kube API and I get a `RestClient::Forbidden (403 Forbidden)`

@mkanoor do you have a service plan and parameters that should work?  I was just doing `ServicePlan.last.order({})` which yielded this payload
```
{
  "apiVersion"=>"servicecatalog.k8s.io/v1beta1",
  "kind"=>"ServiceInstance",
  "metadata"=>{
    "name"=>"eap71-tx-recovery-s2i-e5b1501e-10e7-41bd-814a-884bdda1ec95",
    "namespace"=>"default"
  },
  "spec"=>{
    "clusterServiceClassExternalName"=>"eap71-tx-recovery-s2i",
    "clusterServicePlanExternalName"=>"default",
    "parameters"=>{}
  }
}
```
and these headers
```
{
  "Authorization"=>"Bearer #{TOKEN}",
  "Content-Type"=>"application/json",
  "Accept"=>"application/json"
}
```

app/models/endpoint.rb
```diff
@@ -7,4 +7,8 @@ class Endpoint < ApplicationRecord
   validates :role, :uniqueness => { :scope => :source_id }
 
   acts_as_taggable
+
+  def base_url_path
+    URI::Generic.new(scheme, nil, host, port, nil, path, nil, nil, nil).to_s
```
agrare: `URI::Generic.build` would be more readable since you can pass a hash.  Might just be me but I never liked long multiple positional arguments with mixed in nulls.

E.g. `URI::Generic.build(:scheme => scheme, :host => host, :port => port, :path => path)`

app/models/service_plan.rb
```diff
@@ -12,4 +13,25 @@ class ServicePlan < ApplicationRecord
...
+
+    task = Task.create!(
+      :tenant => tenant,
+      :context => {
+        :service_instance => {
+          :source_id  => source.id,
+          :source_ref => parsed_response['metadata']['selfLink']
```
agrare: We can do this in a follow-up but I'd prefer that we return this to the caller instead of the caller having to know the structure of the kubernetes response.

lib/topological_inventory/core/service_catalog_client.rb
```diff
@@ -0,0 +1,53 @@
...
+      def order_service_plan_url
+        URI.join(@source.default_endpoint.base_url_path, "apis/servicecatalog.k8s.io/v1beta1/namespaces/default/serviceinstances").to_s
+      end
+
+      def make_request(method, url, payload, headers = generic_headers)
+        request_options = {
+          :method     => method,
+          :url        => url,
+          :headers    => headers,
+          :verify_ssl => @source.default_endpoint.verify_ssl,
```
agrare: We made verify_ssl a boolean here instead of an int so you'll have to build the `OpenSSL::SSL::VERIFY_*` yourself.
Something like

`:verify_ssl => @source.default_endpoint.verify_ssl ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE`
We can deal with a custom certificate_authority later

lib/topological_inventory/core/service_catalog_client.rb
```diff
@@ -0,0 +1,53 @@
...
+          plan_name, service_offering_name, catalog_parameters(additional_parameters)
+        )
+        response = make_request(:post, order_service_plan_url, payload)
+        JSON.parse(response.body)
+      end
+
+      private
+
+      def order_service_plan_url
+        URI.join(@source.default_endpoint.base_url_path, "apis/servicecatalog.k8s.io/v1beta1/namespaces/default/serviceinstances").to_s
```
agrare: > `namespaces/default/`

Hm the target namespace should be a parameter right @mkanoor?  Hard-coding to the default-namespace looks like a problem

spec/topological_inventory/core/service_catalog_client_spec.rb
bdunne: @eclarizio You need a v2_key in the root of the project to avoid the error you saw.

lib/topological_inventory/core/service_catalog_client.rb
mkanoor: @agrare 
We are passing the namespace as part of the provider control parameters. The UI is going to have to drop down to pick the namespace.

mkanoor:
@agrare 
For the provisioning the schema has default values, I was using nginx-example

```
{"id":175,"source_id":1,"source_ref":"1e443dec-594a-11e8-ba7e-d094660d31fb","name":"nginx-example","description":"An example Nginx HTTP server and a reverse proxy (nginx) application that serves static content. For more information about using this template, including OpenShift considerations, see https://github.com/sclorg/nginx-ex/blob/master/README.md.","extra":null,"created_at":"2018-10-25T14:26:51.322Z","updated_at":"2018-11-14T22:23:16.272Z","source_deleted_at":null,"tenant_id":1,"source_created_at":"2018-05-16T20:48:37.000Z","archived_on":null,"resource_timestamp":null,"resource_timestamps":{},"resource_timestamps_max":null,"source_region_id":null,"subscription_id":null}
```

You can create a simple JSON payload from this 
https://topological-inventory-api-topological-inventory-ci.10.8.96.54.nip.io/api/v0.0/service_offerings/175/service_plans

lib/topological_inventory/core/service_catalog_client.rb
mkanoor: https://github.com/ManageIQ/service_portal-api/blob/60965b031a61570ca8adf0b0ee6a4c025ca12faa/public/doc/swagger-2.yaml#L526

agrare:
@mkanoor still getting 403 forbidden, going to need you to test this as I'm not sure what is expected?

Here is what I tried:
`Source.first.service_offerings.find_by(:source_ref => '1e443dec-594a-11e8-ba7e-d094660d31fb').service_plans.first.order([{'name' => "NAME", 'value' => "ag_nginx"}, {'name' => "NAMESPACE", 'value' => "topological-inventory-ci"}, {'name' => "NGINX_VERSION", 'value' => "1.13.9"}, {'name' => "MEMORY_LIMIT", 'value' => "100"}, {'name' => "SOURCE_REPOSITORY_URL", 'value' => "abcd"}])`

agrare:
```
You need a hash with these keys filled in
"NAME"
"NAMESPACE"
...
```

It seems like it had to be an array of hashes with name, value keys is that expected?

mkanoor:
@agrare 
It just needs to be a simple hash 
```
{"NAME":"nginx-example-e2c6f44a-6937-4942-ab1f-806defd077ca","NAMESPACE":"openshift","NGINX_VERSION":"1.12","MEMORY_LIMIT":"512Mi","SOURCE_REPOSITORY_URL":"https://github.com/sclorg/nginx-ex.git"}
```

agrare:
@mkanoor that's what I tried first but it blew up [here](https://github.com/ManageIQ/topological_inventory-core/pull/39/files#diff-433892396c24e00281a919318ec93c66R52)

agrare:
And when I try with your hash values I get
```
>> Source.first.service_offerings.find_by(:source_ref => '1e443dec-594a-11e8-ba7e-d094660d31fb').service_plans.first.order([{'name' => "NAME", 'value' => "ag_nginx"}, {'name' => "NAMESPACE", 'value' => "topological-inventory-ci"}, {'name' => "NGINX_VERSION", 'value' => "1.12"}, {'name' => "MEMORY_LIMIT", 'value' => "512Mi"}, {'name' => "SOURCE_REPOSITORY_URL", 'value' => "https://github.com/sclorg/nginx-ex.git"}])
warning: Overriding "Content-Type" header "application/json" with "application/x-www-form-urlencoded" due to payload
       12: from bin/console:15:in `<main>'
       11: from (irb):3
       10: from /home/agrare/src/topological_inventory/topological_inventory-core/app/models/service_plan.rb:18:in `order'
        9: from /home/agrare/src/topological_inventory/topological_inventory-core/lib/topological_inventory/core/service_catalog_client.rb:16:in `order_service_plan'
        8: from /home/agrare/src/topological_inventory/topological_inventory-core/lib/topological_inventory/core/service_catalog_client.rb:35:in `make_request'
        7: from /home/agrare/.gem/gems/rest-client-2.0.2/lib/restclient/request.rb:145:in `execute'
        6: from /home/agrare/.gem/gems/rest-client-2.0.2/lib/restclient/request.rb:715:in `transmit'
        5: from /usr/lib/ruby/2.5.0/net/http.rb:910:in `start'
        4: from /home/agrare/.gem/gems/rest-client-2.0.2/lib/restclient/request.rb:725:in `block in transmit'
        3: from /home/agrare/.gem/gems/rest-client-2.0.2/lib/restclient/request.rb:809:in `process_result'
        2: from /home/agrare/.gem/gems/rest-client-2.0.2/lib/restclient/abstract_response.rb:103:in `return!'
        1: from /home/agrare/.gem/gems/rest-client-2.0.2/lib/restclient/abstract_response.rb:223:in `exception_with_response'
RestClient::Forbidden (403 Forbidden)
```

lib/topological_inventory/core/service_catalog_client.rb
```diff
@@ -0,0 +1,57 @@
...
+          "Content-Type"  => "application/json",
+          "Accept"        => "application/json"
+        }
+      end
+
+      def verify_ssl_mode
+        @source.default_endpoint.verify_ssl ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE
+      end
+
+      def catalog_parameters(parameters)
```
mkanoor: @agrare 
I think this might be the problem, it needs to be a simple hash, in our earlier design we were using an array with name, value, type as separate keys and this was converting it to a simple hash. You should be able to directly use the parameters coming in as long as its a simple hash which looks like this



```
params = {
    "NAME" => "madhu-nginx-example-4000",
    "NAMESPACE" => "openshift",  
    "NGINX_VERSION" => "1.12", 
    "MEMORY_LIMIT" => "512Mi", 
    "SOURCE_REPOSITORY_URL" => "https://github.com/sclorg/nginx-ex.git"
  }

  post_params = {
    "apiVersion" => "servicecatalog.k8s.io/v1beta1",
    "kind"=> "ServiceInstance",
    "metadata" => {
      "name" => service_instance_name,
      "namespace" => namespace
     },
    "spec" => {
       "clusterServiceClassExternalName" => svc_class_name,
       "clusterServicePlanExternalName"  => plan_name,
       "parameters"                      => params
     }
  }
```

lib/topological_inventory/core/service_catalog_client.rb
```diff
@@ -0,0 +1,57 @@
...
+      def initialize(source)
+        @source = source
+      end
+
+      # TODO? Move this to a module in the API repo
+      def order_service_plan(plan_name, service_offering_name, additional_parameters)
+        payload = ServicePlanClient.new.build_payload(
+          plan_name, service_offering_name, catalog_parameters(additional_parameters)
+        )
+        response = make_request(:post, order_service_plan_url, payload)
```
mkanoor: @eclarizio 
payload needs to be converted to payload.to_json

agrare:
Okay changed to using the admin user's token not the management token and after the last two commits I am able to order a service

lib/topological_inventory/core/service_plan_client.rb
```diff
@@ -0,0 +1,21 @@
+module TopologicalInventory
+  module Core
+    class ServicePlanClient
+      def build_payload(service_plan_name, service_offering_name, catalog_parameters)
+        {
+          "apiVersion" => "servicecatalog.k8s.io/v1beta1",
+          "kind"       => "ServiceInstance",
+          "metadata"   => {
+            "name"      => "#{service_offering_name}-#{SecureRandom.uuid}",
+            "namespace" => "default"
```
mkanoor: @agrare The Service Portal can pass in the namespace as part of provider_control_parameters hash. Is this temporary?

spec/topological_inventory/core/service_catalog_client_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        end
+        let(:dummy_response) { {"metadata" => {"selfLink" => "foo"}} }
+        let(:headers) do
+          {
+            "Authorization" => "Bearer token",
+            "Content-Type"  => "application/json",
+            "Accept"        => "application/json"
+          }
+        end
+        let(:catalog_parameters) { [{"name" => "foo", "value" => "bar"}, {"name" => "baz", "value" => "qux"}] }
```
mkanoor: @eclarizio @agrare 
This should just be 
```
{'foo' => 'bar', 'baz' => 'qux'}
```

spec/topological_inventory/core/service_catalog_client_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        let(:service_plan_client) { instance_double("ServicePlanClient") }
+
+        before do
+          allow(ServicePlanClient).to receive(:new).and_return(service_plan_client)
+          allow(service_plan_client).to receive(:build_payload).with(
+            "plan_name", "service_offering_name", {"foo" => "bar", "baz" => "qux"}
```
mkanoor: @eclarizio the hash can be just catalog_parameters

lib/topological_inventory/core/service_plan_client.rb
agrare: That's what I thought too but when I do that I get a 400 bad request

lib/topological_inventory/core/service_catalog_client.rb
agrare: Done here https://github.com/ManageIQ/topological_inventory-core/pull/39/files#diff-0837d2cfce04474e6a42a58c058ad045R17

spec/topological_inventory/core/service_catalog_client_spec.rb
eclarizio: @bdunne Yeah, I'll fix it in a new PR.

PR #41, opened 2018-11-02
Add regions and subscriptions tables and relations

Ladas:


db/migrate/20181102145252_add_regions_and_subscriptions.rb
```diff
@@ -0,0 +1,55 @@
...
+      t.string     "source_ref"
+      t.string     "name"
+      t.timestamps
+      t.datetime   "archived_on", :index => true
+      t.index      %i(source_id source_ref), :unique => true
+    end
+
+    add_foreign_key :subscriptions, :sources, on_delete: :cascade
+
+    create_table "regions", :id => :bigserial do |t|
```
Fryguy: Can we come up with an alternate name?  I'm concerned if/when we bring this back into ManageIQ that this will be a huge problem.

Ladas: @Fryguy @agrare ```source_regions``` maybe?

Fryguy: :+1:

Ladas: and I would guess ```source_subscriptions``` so we keep it consistent?

Fryguy:
I'm ok with subscriptions being subscriptions.  After all, we don't want to add source_ in front of everything.  region just happens to be a special case.

db/migrate/20181102145252_add_regions_and_subscriptions.rb
```diff
@@ -0,0 +1,55 @@
...
+      t.datetime   "archived_on", :index => true
+      t.index      %i(source_id source_ref), :unique => true
+    end
+
+    add_foreign_key :source_regions, :sources, on_delete: :cascade
+    add_foreign_key :source_regions, :subscriptions, on_delete: :cascade
+
+    # Relations to service_offerings
+    add_column :service_offerings, :source_region_id, :bigint
+    add_index :service_offerings, :source_region_id
```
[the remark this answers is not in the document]
Ladas: :+1: ok, since we actually follow the naming convention i can simplify this a lot

Ladas:
@agrare ok simplified and added 1 more change, regions should not be nested to subscriptions, those are 2 top level buckets

PR #42, opened 2018-11-02
Add SSL Security Details to Endpoints

agrare:
Add SSL information to endpoints useful for client connections

### DIFF
--- a/db/migrate/20181102141329_add_ssl_options_to_endpoints.rb
+++ b/db/migrate/20181102141329_add_ssl_options_to_endpoints.rb
@@ -1,0 +1,6 @@
+class AddSslOptionsToEndpoints < ActiveRecord::Migration[5.1]
+  def change
+    add_column :endpoints, :verify_ssl,            :boolean
+    add_column :endpoints, :certificate_authority, :text
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -130,6 +130,8 @@ ActiveRecord::Schema.define(version: 20181102145252) do
     t.string "host"
     t.string "path"
     t.bigint "tenant_id", null: false
+    t.boolean "verify_ssl"
+    t.text "certificate_authority"
     t.index ["source_id"], name: "index_endpoints_on_source_id"
   end
 


db/migrate/20181102141329_add_ssl_options_to_endpoints.rb
```diff
@@ -0,0 +1,7 @@
+class AddSslOptionsToEndpoints < ActiveRecord::Migration[5.1]
+  def change
+    add_column :endpoints, :verify_ssl,            :integer
+    add_column :endpoints, :security_protocol,     :string
```
agrare: @bdunne I added this here but I think scheme, verify_ssl, and certificate_authority cover what we need.  I think security_protocol in MIQ is the equivalent of scheme here, just instead of ssl/non-ssl we have http/https.  Think we need both?

bdunne: I think we should choose `verify_ssl` or `security_protocol`.
Do we need anything other than these:
```ruby
OpenSSL::SSL::VERIFY == :verify_ssl => true,
OpenSSL::SSL::VERIFY_NONE == :verify_ssl => false,
OpenSSL::SSL::VERIFY_PEER == :verify_ssl => true with :certificate_authority,
```

agrare: Nope that should be enough.  I agree I'll stick with just verify_ssl and certificate authority

agrare: Done

db/migrate/20181102141329_add_ssl_options_to_endpoints.rb
```diff
@@ -0,0 +1,6 @@
+class AddSslOptionsToEndpoints < ActiveRecord::Migration[5.1]
+  def change
+    add_column :endpoints, :verify_ssl,            :integer
```
bdunne: Should this be a boolean?

agrare: Oh hm, that adds more logic on the provider to have to check if cert_authority is set because before you could pass verify_ssl in to the native client but I guess it makes it easier on the API

PR #46, opened 2018-11-08
Allow saving of regions and subscriptions

Ladas:


Ladas:
@agrare ping

db/migrate/20181108113645_add_tenant_id_to_regions_and_subscriptions.rb
```diff
@@ -0,0 +1,6 @@
+class AddTenantIdToRegionsAndSubscriptions < ActiveRecord::Migration[5.1]
+  def change
+    add_reference :source_regions, :tenant, :type => :bigint, :index => true, :foreign_key => {:on_delete => :cascade}
+    add_reference :subscriptions, :tenant, :type => :bigint, :index => true, :foreign_key => {:on_delete => :cascade}
```
agrare: These should be `:null => false`

Ladas: right

app/models/source_region.rb
```diff
@@ -0,0 +1,7 @@
+require "archived_concern"
+
+class SourceRegion < ApplicationRecord
+  include ArchivedConcern
+
+  belongs_to :source
```
agrare: These should belong to `:tenant` and `Tenant` should have many regions and subscriptions

agrare:
Just the one thing then LGTM

app/models/source_region.rb
Ladas: done

PR #47, opened 2018-11-08
Add container model

agrare:
Add a container model belonging to a container_group and store the
memory and cpu limits/requests


db/schema.rb
```diff
@@ -225,7 +247,7 @@
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
     t.datetime "archived_on"
-    t.bigint "tenant_id"
+    t.bigint "tenant_id", null: false
```
Ladas: hm, I didn't push these in my previous migration (we are missing the check)

db/schema.rb
```diff
@@ -119,6 +119,28 @@
...
+    t.jsonb "resource_timestamps", default: {}
+    t.datetime "resource_timestamps_max"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.datetime "source_created_at"
+    t.datetime "source_deleted_at"
+    t.datetime "archived_on"
+    t.index ["archived_on"], name: "index_containers_on_archived_on"
+    t.index ["container_group_id", "name"], name: "index_containers_on_container_group_id_and_name", unique: true
+    t.index ["container_group_id"], name: "index_containers_on_container_group_id"
```
Ladas: I think this index might be redundant, the unique index should be used for query ```.where(:container_group_id => XY)```

db/schema.rb
agrare: Yeah we need to re-run db:migrate to write out the db/schema.rb file after changing a migration

db/schema.rb
agrare: :+1:

PR #50, opened 2018-11-09
Seed the "openshift" SourceType

bdunne:


agrare:
@bdunne we need to define a seed class, https://github.com/ManageIQ/topological_inventory-core/blob/master/Rakefile#L43-L45 since something is required and https://github.com/rails/rails/blob/master/activerecord/lib/active_record/tasks/database_tasks.rb#L111 falls back to `Rails.application`

agrare:
Maybe `TopologicalInventory::Core::SeedLoader` ?

Fryguy:
Still not a fan of core repo knowing anything about any particular "plugin"

bdunne:
> Still not a fan of core repo knowing anything about any particular "plugin"

I'm open to alternatives...

bdunne:
I'm planning to enumerate over the `SourceTypes` and their `Sources` to deploy the collectors.

agrare:
@Fryguy I think the difference here is that we won't have any openshift specific code like we do in the MIQ plugins, it is just the type.  Having an repo for each type with just a seed file seems like overkill but we can do that.

lib/topological_inventory/persister/ar_helper.rb
```diff
@@ -2,20 +2,38 @@ module TopologicalInventory
   module Persister
     module ArHelper
       def self.load_environment!
+        ENV["RAILS_ENV"] ||= "development"
+
+        database_yaml_path = root.join("config", "database.yml")
+        database_yaml = YAML.load_file(database_yaml_path) if database_yaml_path.exist?
```
agrare: Need to require yaml here and we can probably remove this now https://github.com/ManageIQ/topological_inventory-core/blob/5dc481e54a98ebdf0a8109237efb91b700bec88d/Rakefile#L23

lib/topological_inventory/persister/ar_helper.rb
```diff
@@ -2,20 +2,38 @@ module TopologicalInventory
...
+
+        ActiveRecord::Tasks::DatabaseTasks.env = ENV["RAILS_ENV"]
```
agrare: Need to require active_record here too

```
	1: from bin/topological_inventory-persister:10:in `<main>'
/home/grare/adam/src/topological_inventory/topological_inventory-core/lib/topological_inventory/persister/ar_helper.rb:11:in `load_environment!': uninitialized constant TopologicalInventory::Persister::ArHelper::ActiveRecord (NameError)
```

PR #51, opened 2018-11-13
Add VMs and OrchestrationStacks for AWS refresh

agrare:
Adds models for Vms and OrchestrationStacks for AWS inventory.

I left out a `:template` from the vms table with the theory that we'll have separate tables for vms and templates.

### DIFF
--- a/db/migrate/20181113182615_add_vms_table.rb
+++ b/db/migrate/20181113182615_add_vms_table.rb
@@ -5,7 +5,7 @@ class AddVmsTable < ActiveRecord::Migration[5.1]
       t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
 
       t.string :source_ref
-      t.string :uuid
+      t.string :uid_ems
       t.string :name
       t.string :hostname
       t.string :description
@@ -26,7 +26,7 @@ class AddVmsTable < ActiveRecord::Migration[5.1]
 
       t.index [:source_id, :source_ref], :unique => true
       t.index :archived_on
-      t.index :uuid
+      t.index :uid_ems
     end
   end
 end
--- a/db/migrate/20181113182615_add_vms_table.rb
+++ b/db/migrate/20181113182615_add_vms_table.rb
@@ -13,6 +13,8 @@ class AddVmsTable < ActiveRecord::Migration[5.1]
       t.bigint :cpus
       t.bigint :memory
 
+      t.jsonb :extra
+
       t.datetime :resource_timestamp
       t.jsonb    :resource_timestamps, default: {}
       t.datetime :resource_timestamps_max
--- a/db/migrate/20181113182615_add_vms_table.rb
+++ b/db/migrate/20181113182615_add_vms_table.rb
@@ -1,0 +1,26 @@
+class AddVmsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :vms do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string :source_ref
+      t.string :uuid
+      t.string :vendor
+      t.string :name
+      t.string :hostname
+      t.string :description
+      t.string :power_state
+      t.bigint :cpus
+      t.bigint :memory
+
+      t.timestamps
+      t.datetime :archived_at
+      t.datetime :source_created_at
+      t.datetime :source_deleted_at
+
+      t.index [:source_id, :source_ref], :unique => true
+      t.index :archived_at
+    end
+  end
+end
--- a/db/migrate/20181113182615_add_vms_table.rb
+++ b/db/migrate/20181113182615_add_vms_table.rb
@@ -14,6 +14,10 @@ class AddVmsTable < ActiveRecord::Migration[5.1]
       t.bigint :cpus
       t.bigint :memory
 
+      t.datetime :resource_timestamp
+      t.jsonb    :resource_timestamps, default: {}
+      t.datetime :resource_timestamps_max
+
       t.timestamps
       t.datetime :archived_at
       t.datetime :source_created_at
--- a/db/migrate/20181113182615_add_vms_table.rb
+++ b/db/migrate/20181113182615_add_vms_table.rb
@@ -6,7 +6,6 @@ class AddVmsTable < ActiveRecord::Migration[5.1]
 
       t.string :source_ref
       t.string :uuid
-      t.string :vendor
       t.string :name
       t.string :hostname
       t.string :description
--- a/db/migrate/20181113182615_add_vms_table.rb
+++ b/db/migrate/20181113182615_add_vms_table.rb
@@ -26,6 +26,7 @@ class AddVmsTable < ActiveRecord::Migration[5.1]
 
       t.index [:source_id, :source_ref], :unique => true
       t.index :archived_on
+      t.index :uuid
     end
   end
 end
--- a/db/migrate/20181113182615_add_vms_table.rb
+++ b/db/migrate/20181113182615_add_vms_table.rb
@@ -20,12 +20,12 @@ class AddVmsTable < ActiveRecord::Migration[5.1]
       t.datetime :resource_timestamps_max
 
       t.timestamps
-      t.datetime :archived_at
+      t.datetime :archived_on
       t.datetime :source_created_at
       t.datetime :source_deleted_at
 
       t.index [:source_id, :source_ref], :unique => true
-      t.index :archived_at
+      t.index :archived_on
     end
   end
 end
--- a/db/migrate/20181113190507_add_orchestration_stacks.rb
+++ b/db/migrate/20181113190507_add_orchestration_stacks.rb
@@ -13,12 +13,12 @@ class AddOrchestrationStacks < ActiveRecord::Migration[5.1]
       t.datetime :resource_timestamps_max
 
       t.timestamps
-      t.datetime :archived_at
+      t.datetime :archived_on
       t.datetime :source_created_at
       t.datetime :source_deleted_at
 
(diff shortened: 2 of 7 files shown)


db/migrate/20181113182615_add_vms_table.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.bigint :cpus
+      t.bigint :memory
+
+      t.timestamps
+      t.datetime :archived_at
+      t.datetime :source_created_at
+      t.datetime :source_deleted_at
+
+      t.index [:source_id, :source_ref], :unique => true
+      t.index :archived_at
```
Ladas: should be ```archived_on```

db/migrate/20181113182615_add_vms_table.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.string :hostname
+      t.string :description
+      t.string :power_state
+      t.bigint :cpus
+      t.bigint :memory
+
+      t.timestamps
+      t.datetime :archived_at
+      t.datetime :source_created_at
+      t.datetime :source_deleted_at
```
Ladas: and lets also add 
```
t.datetime "resource_timestamp"
t.jsonb "resource_timestamps", default: {}
t.datetime "resource_timestamps_max"
```

db/migrate/20181113182615_add_vms_table.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string :source_ref
+      t.string :uuid
+      t.string :vendor
+      t.string :name
+      t.string :hostname
+      t.string :description
+      t.string :power_state
+      t.bigint :cpus
```
Ladas: is this enough? I think we also have :virtual_cpus and :sockets?

db/migrate/20181113182615_add_vms_table.rb
```diff
@@ -0,0 +1,26 @@
+class AddVmsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :vms do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string :source_ref
+      t.string :uuid
+      t.string :vendor
```
Ladas: isn't the source_type same thing as :vendor?

db/migrate/20181113182615_add_vms_table.rb
```diff
@@ -0,0 +1,26 @@
+class AddVmsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :vms do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string :source_ref
+      t.string :uuid
```
Ladas: will this be external_uuid or internal?

db/migrate/20181113190507_add_orchestration_stacks.rb
```diff
@@ -0,0 +1,22 @@
+class AddOrchestrationStacks < ActiveRecord::Migration[5.1]
+  def change
+    create_table :orchestration_stacks do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
```
Ladas: we should also have ```parent_orchestration_id```

I think references won't allow you to add it with renamed column, so you'll need to add the column and index manually and:

```
add_foreign_key :orchestration_stacks, :orchestration_stacks, on_delete: :cascade, column: 'parent_orchestration_stack_id'
```

db/migrate/20181113190507_add_orchestration_stacks.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.string :name
+      t.string :description
+
+      t.timestamps
+      t.datetime :archived_at
+      t.datetime :source_created_at
+      t.datetime :source_deleted_at
+
+      t.index [:source_id, :source_ref], :unique => true
+      t.index :archived_at
```
Ladas: also :archived_on

db/migrate/20181113190507_add_orchestration_stacks.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    add_reference :vms, :orchestration_stack, :index => true, :foreign_key => {:on_delete => :nullify}
```
Ladas: I think we'll want this on :orchestration_stack_resource, since it's the resource that maps relation to other entity. And I am not sure what'll be the best here, since resource can tie to any model in the system. Will we want foreign_key per each? (if we'd go with generic object, maybe it's not that bad) or would we go with N:M table with polymorphic relation? (but then we can't do foreign_key constraint)

db/migrate/20181113182615_add_vms_table.rb
agrare: Yeah I saw that on the other models and I think we should change them to archived_at.  Every other timestamp is `_at`:
`created_at`, `updated_at`, `source_created_at`, and `source_deleted_at`

db/migrate/20181113182615_add_vms_table.rb
agrare: External, thinking the replacement for uid_ems.  Might just call it uid_ems, trying to keep the naming closer to MIQ than we've been doing.

db/migrate/20181113182615_add_vms_table.rb
agrare: yeah good point

db/migrate/20181113182615_add_vms_table.rb
agrare: Yeah I figured that is mostly an infra vm thing and we could put in the extra jsonb (which I should add)

agrare: Okay added a jsonb extra column

db/migrate/20181113190507_add_orchestration_stacks.rb
agrare: Is that for a parent stack or a polymorphic belongs_to resource?

db/migrate/20181113190507_add_orchestration_stacks.rb
agrare: Hm okay is that different than MIQ then?  MIQ a vm belongs_to orchestration_stack

db/migrate/20181113190507_add_orchestration_stacks.rb
Ladas: that would be for the parent stack

db/migrate/20181113190507_add_orchestration_stacks.rb
Ladas: ah, ok, I was looking at this the other way around.

Ok, we can probably keep it this way

db/migrate/20181113182615_add_vms_table.rb
Ladas: right, we've talked about this. Ok I am fine with :uid_ems, or :external_uuid

We might have cases where compose ems_ref from multiple attributes though, so maybe this could just be in :extra column?

agrare: I think we'll use this for cross-linking right?  I can name it uid_ems but I think having it separate is a good idea since this is a common concept.

db/migrate/20181113190507_add_orchestration_stacks.rb
agrare: I don't see this on MIQ so I'll leave it out and we can revisit in a future PR if we need it

agrare:
@bdunne can you take a look?

db/migrate/20181113182615_add_vms_table.rb
```diff
@@ -0,0 +1,32 @@
...
+    create_table :vms do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string :source_ref
+      t.string :uid_ems
+      t.string :name
+      t.string :hostname
+      t.string :description
+      t.string :power_state
```
bdunne: Do we also need `raw_power_state`?

db/migrate/20181113182615_add_vms_table.rb
```diff
@@ -0,0 +1,32 @@
+class AddVmsTable < ActiveRecord::Migration[5.1]
+  def change
+    create_table :vms do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string :source_ref
+      t.string :uid_ems
```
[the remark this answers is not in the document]
agrare: I'm trying to keep the column names the same as MIQ :)

db/migrate/20181113182615_add_vms_table.rb
agrare: Yeah I left that out because that sounds like something the parser should just set correctly, I don't know why we have both columns

bdunne: According to https://github.com/ManageIQ/manageiq/pull/481 it seems the action to take on the VM is determined by the power state from the provider.

agrare: Looks like that is something that could use the `:extra` column on a per-provider basis e.g. openstack could store its raw_power_state there but we should never need to know that at the normalized level.

PR #52, opened 2018-11-14
Add name indexes to node and namespace

Ladas:
We are using name for secondary index, so we need db index to
speedup the lookup.


Ladas:
@agrare cc

PR #54, opened 2018-11-15
Add models for mark and sweep

Ladas:
Add models for mark and sweep and for sweeping job, we are allowing to re-queue it, if all the parts of the refresh were not finished.

### DIFF
--- a/db/migrate/20181114094606_add_last_seen_at_to_all_tables.rb
+++ b/db/migrate/20181114094606_add_last_seen_at_to_all_tables.rb
@@ -1,0 +1,36 @@
+class AddLastSeenAtToAllTables < ActiveRecord::Migration[5.1]
+  def change
+    add_column :container_groups, :last_seen_at, :datetime
+    add_index :container_groups, :last_seen_at
+
+    add_column :container_nodes, :last_seen_at, :datetime
+    add_index :container_nodes, :last_seen_at
+
+    add_column :container_projects, :last_seen_at, :datetime
+    add_index :container_projects, :last_seen_at
+
+    add_column :container_templates, :last_seen_at, :datetime
+    add_index :container_templates, :last_seen_at
+
+    add_column :service_instances, :last_seen_at, :datetime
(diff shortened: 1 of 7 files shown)


db/migrate/20181114094606_add_last_seen_at_to_all_tables.rb
```diff
@@ -0,0 +1,36 @@
+class AddLastSeenAtToAllTables < ActiveRecord::Migration[5.1]
+  def change
+    add_column :container_groups, :last_seen_at, :datetime
+    add_index :container_groups, :last_seen_at
```
agrare: Pretty sure you can just pass `:index => true` to `add_column`

agrare:
@Ladas are we going to clear these states after a successful sweep?  These seem like they could build up really quickly.

Ladas:
@agrare I think we'll just connect it to the cleanup worker (cleaning up also old archived items, etc.)

Cause at least for the beginning, this actually shows us if all went fine (before we have a common logging in and nice stats from that)

db/migrate/20181114094606_add_last_seen_at_to_all_tables.rb
Ladas: ah, looks like you can't do that :-)

Ladas:
@agrare hm, I could probably delete the :finished parts, as part of sweeping

db/migrate/20181114094606_add_last_seen_at_to_all_tables.rb
agrare: Hm okay, we do it here https://github.com/ManageIQ/manageiq-schema/blob/master/db/migrate/20180920085721_add_maintenance_zone_id_to_region.rb#L3
:worried:

Ladas: ah, right, yeah it's funky, it doesn't throw exception, but it doesn't add the index either :-)

PR #55, opened 2018-11-19
Vm#orchestration_stack is an optional parameter

agrare:
Allow vm#orchestration_stack to be null matching the database schema.

### DIFF
--- a/app/models/vm.rb
+++ b/app/models/vm.rb
@@ -5,7 +5,7 @@ class Vm < ApplicationRecord
 
   belongs_to :tenant
   belongs_to :source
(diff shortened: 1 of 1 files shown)


app/models/vm.rb
```diff
@@ -5,7 +5,7 @@ class Vm < ApplicationRecord
 
   belongs_to :tenant
   belongs_to :source
-  belongs_to :orchestration_stack
+  belongs_to :orchestration_stack, :optional => true
```
Ladas: ah, could we maybe disable these? We have not null in the DB, so we know what is mandatory

PR #57, opened 2018-11-19
Add VM cross linking

agrare:
Test cross linking support


lib/topological_inventory/schema/default.rb
```diff
@@ -51,12 +54,27 @@ def add_secondary_refs_name(builder)
...
         add_collection(:containers) do |builder|
           add_default_properties(builder, manager_ref: [:container_group, :name])
           builder.add_default_values(:tenant_id => ->(persister) { persister.manager.tenant_id })
         end
       end
+
+      def add_cross_link_vm
+        add_collection(:cross_link_vm) do |builder|
+          builder.add_properties(
+            :arel           => Vm,
```
Ladas: ```Vm.where(:tenant_id => manager.tenant_id)```

I think manager is accessible here, right?

agrare: Oh yeah good catch

lib/topological_inventory/schema/default.rb
```diff
@@ -51,12 +54,27 @@ def add_secondary_refs_name(builder)
...
+            :model_class    => Vm,
```
Ladas: and lets define unique name

```:cross_link_vms``` and aglign it with ```add_cross_link_vms```

Comment (Ladas) [reply lib/topological_inventory/schema/default.rb]: also we will need 
```
:strategy => :local_db_find_references # so it will just search the db and not store anything
```

agrare:
@bdunne can you take a look?

agrare:
Thanks @bdunne

PR #58, opened 2018-11-22
Mark and sweep control mechanism

Ladas:
Mark and sweep control mechanism

Depends on:
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/54
- [x]  https://github.com/ManageIQ/topological_inventory-ingress_api-client-ruby/pull/9

### DIFF
--- a/lib/topological_inventory/persister/workflow.rb
+++ b/lib/topological_inventory/persister/workflow.rb
@@ -89,7 +89,7 @@ module TopologicalInventory
       def build_refresh_state_part(refresh_states_ic, refresh_state_parts_ic, status, error_message)
         return unless status
 
-        refresh_state_part_data                 = {
+        refresh_state_part_data = {
           :uuid          => refresh_state_part_uuid,
           :refresh_state => refresh_states_ic.lazy_find(:uuid => refresh_state_uuid),
           :status        => status
--- a/lib/topological_inventory/persister/workflow.rb
+++ b/lib/topological_inventory/persister/workflow.rb
@@ -1,0 +1,182 @@
+require "topological_inventory/persister/logging"
+
+module TopologicalInventory
+  module Persister
+    class Workflow
+      include Logging
+
+      def initialize(persister, messaging_client, payload)
+        @persister        = persister
+        @messaging_client = messaging_client
+        # TODO(lsmola) we should be able to reconstruct the payload out of persistor? E.g. to just repeat the sweep phase
+        @payload = payload
+      end
+
+      def execute!
+        persist_collections!
+
+        if total_parts
+          sweep_inactive_records!
+        end
+      end
+
+      private
+
+      attr_reader :persister, :messaging_client
+      delegate :manager,
+               :persist!,
+               :refresh_state_uuid,
+               :refresh_state_part_uuid,
+               :total_parts,
+               :sweep_scope,
+               :to => :persister
+
+      def define_refresh_state_ics
+        refresh_states_inventory_collection = ::InventoryRefresh::InventoryCollection.new(
+          :manager_ref                 => [:uuid],
+          :saver_strategy              => :concurrent_safe_batch,
+          :parent                      => manager,
+          :association                 => :refresh_states,
+          :create_only                 => true,
+          :model_class                 => RefreshState,
+          :inventory_object_attributes => %i(ems_id uuid status source_id tenant_id),
+          :default_values              => {
+            :source_id => manager.id,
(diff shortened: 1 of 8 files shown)


Ladas:
cc @agrare

lib/topological_inventory/persister/workflow.rb
```diff
@@ -0,0 +1,183 @@
...
+               :manager,
+               :persist!,
+               :refresh_state_uuid,
+               :refresh_state_part_uuid,
+               :total_parts,
+               :sweep_scope,
+               :to => :persister
+
+      def define_refresh_state_ics
+        refresh_states_inventory_collection = ::InventoryRefresh::InventoryCollection.new(
```
agrare: These should just be defined in https://github.com/ManageIQ/topological_inventory-core/tree/master/lib/topological_inventory/schema/default.rb

Ladas: ah, right, these should not be part of refresh. I use these because I don't have upsert query abstracted enough. And I need to do upsert, so this works.

agrare: Oh okay so they won't be inventory collections long term?

Ladas: yes, these should be just 2 upserts

lib/topological_inventory/persister/workflow.rb
```diff
@@ -0,0 +1,183 @@
...
+      def initialize(persister, messaging_client, payload)
+        @persister        = persister
+        @messaging_client = messaging_client
+        # TODO(lsmola) we should be able to reconstruct the payload out of persistor? E.g. to just repeat the sweep phase
+        @payload = payload
+      end
+
+      def execute!
+        if total_parts
+          sweep_inactive_records!
```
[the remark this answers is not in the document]
Ladas: That could theoretically work, but there would be more edge cases to cover. E.g. after we set total parts, all parts are already saved, so the sweeping would not be performed. Or 2 or more jobs will figure out ```total_parts == refresh_state.refresh_state_parts.count``` and do the sweeping, so we'll do extra queries and notify about deleted objects twice.

lib/topological_inventory/persister/workflow.rb
```diff
@@ -0,0 +1,183 @@
...
+
+      attr_reader :persister, :messaging_client, :payload
+      delegate :inventory_collections,
+               :manager,
+               :persist!,
+               :refresh_state_uuid,
+               :refresh_state_part_uuid,
+               :total_parts,
+               :sweep_scope,
+               :to => :persister
```
agrare: This looks like it would be cleaner if this wasn't in a subclass but instead in the persister class.  We wouldn't need to pass the client in, we wouldn't need to delegate all this up, etc...

lib/topological_inventory/persister/workflow.rb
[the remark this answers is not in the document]
agrare: > after we set total parts, all parts are already saved, so the sweeping would not be performed. 

Well that's an easy one, do the sweep_inactive_records check for everything not just the inventory chunks, the last total_parts wouldn't even have to be its own message type.

> Or 2 or more jobs will figure out total_parts == refresh_state.refresh_state_parts.count and do the sweeping, so we'll do extra queries and notify about deleted objects twice.

There's really no way to do the check atomically to ensure that only one caller does this?  Feels like something we do a lot already.

lib/topological_inventory/persister/workflow.rb
Ladas: hm, the persister class is big enough already? The workflow process feels separate

lib/topological_inventory/persister/workflow.rb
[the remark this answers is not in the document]
Ladas: > > after we set total parts, all parts are already saved, so the sweeping would not be performed.
> 
> Well that's an easy one, do the sweep_inactive_records check for everything not just the inventory chunks, the last total_parts wouldn't even have to be its own message type.
> 

Hm not sure I follow, now the check is always tied to one refresh uuid.

> > Or 2 or more jobs will figure out total_parts == refresh_state.refresh_state_parts.count and do the sweeping, so we'll do extra queries and notify about deleted objects twice.
> 
> There's really no way to do the check atomically to ensure that only one caller does this? Feels like something we do a lot already.

Well we could do locking, but it feels like dedicated job is just cleaner (also, all the current jobs can run twice, if you restart the worker in the middle of the job. )

PR #59, opened 2018-11-23
Schema to_hash method

Ladas:
Schema to_hash method

### DIFF
--- a/lib/topological_inventory/schema/default.rb
+++ b/lib/topological_inventory/schema/default.rb
@@ -37,6 +37,15 @@ module TopologicalInventory
         add_cross_link_vms
       end
 
+      def to_hash
+        {
+          "schema": {
+            "name": "Default"
+          },
+          "source": manager.uid,
+        }.merge(super)
+      end
+
       def targeted?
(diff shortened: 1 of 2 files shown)


lib/topological_inventory/schema/default.rb
```diff
@@ -37,6 +37,15 @@ def initialize_inventory_collections
         add_cross_link_vms
       end
 
+      def to_hash
+        {
+          "schema": {
+            "name": "Default"
```
agrare: Can we get the schema name dynamically that way if we have other schemas that inherit from this one they can just the same method?

Ladas: I think we can do that once we have some ind of base schema?

Ladas: or should I just put that to persister?

agrare: No I think here makes sense, so can you add a `TopologicalInventory::Schema::Base` with this there?  Not sure if any of the other helper methods make sense in base but some might.

PR #60, opened 2018-11-27
Add models and schema for container images

agrare:
Add models for ContainerImages


db/migrate/20181127191451_add_container_images.rb
```diff
@@ -0,0 +1,28 @@
...
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string     :source_ref
+      t.string     :resource_version
+      t.string     :name
+      t.string     :tag
+
+      t.timestamps
+      t.datetime   :archived_on
```
bdunne: Should this be `:archived_at` to match the rest of the timestamp columns?

agrare: Yeah I completely agree I've been talking about this with @Ladas I think we need to make these consistent but right now inventory_refresh gem assumes archived_on and we need to change all of these at once.

PR #61, opened 2018-11-28
Add flavor model

Ladas:
Add flavor model


Ladas:
cc @agrare

agrare:
The hakiri warnings don't look related to this PR, I'll see if there are newer versions of AR that we could update to to fix these

PR #62, opened 2018-11-28
Add json schema to source_types for source create

agrare:
Add a json schema for the form for source creation.


agrare:
@martinpovolny Fields for the openshift source type should be:

```
:source
  :name: (string) required
  :endpoints:
  - :role: kubernetes
    :scheme: (string) required default 'https'
    :host: (string) required
    :port: (integer) required default 8443
    :verify_ssl: (boolean) required default true
    :certificate_authority: (string) optional if verify_ssl is true, not allowed if verify_ssl is false
    :authentications:
    - :token: (string) required
```

martinpovolny:
Thx!

PR #64, opened 2018-11-28
 Move ArHelper to core

Fryguy:
The ArHelper is not specific to the persister, so should be part of core.  Additionally, getting it out of the way simplifies extraction of the persister.


Fryguy:
@agrare Please review.

PR #67, opened 2018-11-29
[WIP] Add core tagging inventory collection and tests

agrare:
~~Horrible, terrible, really bad hack just to start getting tests and collectors working.~~ This is now less terrible than it was previously.

### DIFF
--- a/app/models/tagging.rb
+++ b/app/models/tagging.rb
@@ -1,0 +1,6 @@
+class Tagging < ApplicationRecord
+  belongs_to :taggable, :polymorphic => true
+  belongs_to :tagger,   :polymorphic => true
+
+  belongs_to :tag
+end
--- a/app/models/tagging.rb
+++ b/app/models/tagging.rb
@@ -1,4 +1,7 @@
 class Tagging < ApplicationRecord
+  belongs_to :tenant
+  belongs_to :source
+
   belongs_to :taggable, :polymorphic => true
   belongs_to :tagger,   :polymorphic => true
 
--- a/lib/topological_inventory/schema/default.rb
+++ b/lib/topological_inventory/schema/default.rb
@@ -31,7 +31,7 @@ module TopologicalInventory
       private
 
       def add_default_collection(model)
-        add_collection(model, InventoryRefresh::InventoryCollection::Builder, :attributes_blacklist => [:tags]) do |builder|
+        add_collection(model) do |builder|
           add_default_properties(builder)
           add_default_values(builder)
           yield builder if block_given?
@@ -82,7 +82,10 @@ module TopologicalInventory
         add_collection(:taggings) do |builder|
           builder.add_properties(
             :manager_ref => [:taggable, :tag],
+            :strategy           => :local_db_find_missing_references,
+            :saver_strategy     => :concurrent_safe_batch,
           )
+          add_default_values(builder)
         end
       end
 
@@ -90,8 +93,10 @@ module TopologicalInventory
         add_collection(:tags) do |builder|
           builder.add_properties(
             :manager_ref => [:name],
-            :parent_inventory_collections => [:taggings],
+            :strategy           => :local_db_find_missing_references,
+            :saver_strategy     => :concurrent_safe_batch,
           )
+          add_default_values(builder)
         end
       end
     end
--- a/lib/topological_inventory/schema/default.rb
+++ b/lib/topological_inventory/schema/default.rb
@@ -6,6 +6,7 @@ module TopologicalInventory
     class Default < InventoryRefresh::Persister
       def initialize_inventory_collections
         add_containers
+        add_cross_link_vms
         add_default_collection(:container_groups)
         add_default_collection(:container_images)
         add_default_collection(:container_nodes)    { |b| add_secondary_refs_name(b) }
@@ -19,7 +20,7 @@ module TopologicalInventory
         add_default_collection(:source_regions)
         add_default_collection(:subscriptions)
         add_default_collection(:vms)
-        add_cross_link_vms
+        add_taggings
         add_tags
       end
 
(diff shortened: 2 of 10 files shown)


agrare:
@Ladas this is gross but I don't see a way of doing bulk assignment with `acts-as-taggable-on` so we might have to write our own.

lib/topological_inventory/schema/default.rb
```diff
@@ -20,6 +20,7 @@ def initialize_inventory_collections
         add_default_collection(:subscriptions)
         add_default_collection(:vms)
         add_cross_link_vms
+        add_tags
```
Ladas: so we should be able to create records in taggings table directly so modeling tagging IC? Adding also the tags IC.

In OpenShift we did taggings per each parent IC like  https://github.com/Ladas/manageiq-providers-kubernetes/blob/9eb61a16c5cdbbe22b72fb3de74e173cdab32935/app/models/manageiq/providers/kubernetes/inventory/persister/definitions/container_collections.rb#L176

Then we can just do:
```ruby
collections['taggings_for_vms'].build(:tag => collections['tags].lazy_find('super_tag'), :taggable => collections['vms'].lazy_find(XY))
```

And you don't need to worry about ```:tag => collections['tags].lazy_find('super_tag')```, since we just have :name in that table, so we'll just do skeletal precreate. (not sure why is there ```tagging_count``` in that table, that looks unrelated to multi tenant world, we should just delete that column)

Ladas: we'll need to enhance the unique indexes though https://github.com/ManageIQ/topological_inventory-core/blob/master/db/schema.rb#L411

look e.g. at hardwares here https://github.com/ManageIQ/manageiq-schema/pull/19/files#diff-5214f0a7b9057034ce42bbc155d56284R86

in the current setup, you'll need also an index that where the tagger cols and context are nil? We'll not be filling those, correct? If that is the case, maybe we should delete these cols?

spec/persister/test_inventory/tags_inventory.json
```diff
@@ -0,0 +1 @@
+{"name":"OCP","schema":{"name":"Default"},"source":"9a874712-9a55-49ab-a46a-c823acc35503","collections":[{"name":"service_offerings","data":[{"source_ref":"afca9dde-0888-412f-bc81-383f2fadbce5","name":"postgresql","tags":["database","postgresql"]}]}]}
```
Ladas: can we have this formatted nicely? Does you editor have autoformat?

lib/topological_inventory/schema/default.rb
Ladas: we want to create ```tags``` dynamically, right? Btw. the other alternative would be to store in into jsonb, I could check if we can index those. Wouldn't that be better?

app/models/tagging.rb
```diff
@@ -0,0 +1,6 @@
+class Tagging < ApplicationRecord
+  belongs_to :taggable, :polymorphic => true
+  belongs_to :tagger,   :polymorphic => true
```
Ladas: @agrare do we need the tagger? Is this for some future use?

agrare: It was in the table so I added an assoc for it.  Since we prob won't have our users in our db we can probably drop it from the table if we get rid of acts-as-taggable-on.

lib/topological_inventory/schema/default.rb
agrare: Hm why would we need a different IC per parent if we can just set the taggable type/id?  That would make it easier to tag anything without having double the ICs (one for the actual collection and one for the taggigns)?

agrare: > (not sure why is there tagging_count in that table, that looks unrelated to multi tenant world, we should just delete that column)

Yeah agreed, I want to leave the schema the same though until we decide if we're dropping acts-as-taggable-on or just not using it to set tags from inventory refresh.

agrare: > we want to create tags dynamically, right?

Right

> Btw. the other alternative would be to store in into jsonb, I could check if we can index those. Wouldn't that be better?

Possibly?  It depends on how we're using the tags, would be nice to get this working then when we know how we're going to use the tags on the calling side do some perf tests.

agrare: > so we should be able to create records in taggings table directly so modeling tagging IC? Adding also the tags IC.

Oh yeah, and I did add both taggings and tags ICs https://github.com/ManageIQ/topological_inventory-core/pull/67/files#diff-bf581c67a3bfa3b364180f998a222a83R23

Ladas: > Hm why would we need a different IC per parent if we can just set the taggable type/id? That would make it easier to tag anything without having double the ICs (one for the actual collection and one for the taggigns)?

The query is build from source. So e.g. source joins VM that joins tagging (where the relation has the right type baked in, so we can do the join)

If you'd want to do 1 query for e.g. VM and Network, we'd need to do union of 2 queries. And that query would get expensive, for all taggable objects.

Using just taggable type/id for filtering. Right now, we have no way do define the query, to ensure it stays under 1 scope (well we could rely on the fact the lazy_finds are bound to the scope). And no way to e.g. build a query to give you the all tags of list of entities, so you can delete the inactive ones. Well we'd have to build custom saver to attempt this.

lib/topological_inventory/schema/default.rb
```diff
@@ -75,6 +77,23 @@ def add_cross_link_vms
           )
         end
       end
+
+      def add_taggings
+        add_collection(:taggings) do |builder|
+          builder.add_properties(
+            :manager_ref => [:taggable],
```
Ladas: ```[:taggable, :tag``` ?

agrare: :+1: yeah missed that

lib/topological_inventory/schema/default.rb
```diff
@@ -75,6 +77,23 @@ def add_cross_link_vms
...
+          )
+        end
+      end
+
+      def add_tags
+        add_collection(:tags) do |builder|
+          builder.add_properties(
+            :manager_ref => [:tagging, :name],
```
Ladas: just ```:manager_ref => [:name],```

PR #68, opened 2018-11-30
Schema and models for storage

Ladas:


db/schema.rb
```diff
@@ -175,6 +175,20 @@
...
+    t.bigint "tenant_id", null: false
+    t.bigint "vm_id", null: false
+    t.bigint "volume_id"
+    t.string "device"
+    t.string "size"
+    t.string "status"
+    t.jsonb "extra"
+    t.index ["tenant_id"], name: "index_disks_on_tenant_id"
+    t.index ["vm_id", "device"], name: "index_disks_on_vm_id_and_device", unique: true
+    t.index ["vm_id"], name: "index_disks_on_vm_id"
```
Ladas: @agrare hm, we should eliminate all indexes like this, the uniq column should already cover this.

Ladas:
@agrare it's a draft, but i'd like a feedback

db/migrate/20181130111723_add_volumes.rb
```diff
@@ -0,0 +1,61 @@
...
+      t.index :archived_on
+    end
+
+    create_table :disks, :id => :bigserial do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :vm, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :volume, :type => :bigint, :index => true, :foreign_key => {:on_delete => :cascade}
+
+      t.string :device
+      t.string :size
```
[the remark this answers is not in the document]
Ladas: yes, not sure if I like it though. Maybe we could separate local disks from attachments?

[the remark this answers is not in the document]
agrare: Yeah maybe...we could have disks and cloud_volumes be different since standard disks probably map really well to the infra side also.

db/migrate/20181130111723_add_volumes.rb
```diff
@@ -0,0 +1,61 @@
+class AddVolumes < ActiveRecord::Migration[5.1]
+  def change
+    create_table :volumes, :id => :bigserial do |t|
```
agrare: Wonder if we should use `cloud_volumes` ?  Just volumes is a little generic and with the attachments to vms I don't want to get into a situation like hardwares where its so generic it belongs to 3 different things

db/migrate/20181130111723_add_volumes.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      t.datetime   :resource_timestamp
+      t.jsonb      :resource_timestamps, :default => {}
+      t.datetime   :resource_timestamps_max
+
+      t.index [:source_id, :source_ref], :unique => true
+      t.index :archived_on
+    end
+
+    create_table :volume_types, :id => :bigserial do |t|
```
agrare: @Ladas what are some examples of volume types?  Is this something that could be a string property on a volume or is it really a managed inventory item?  And do we need volume_types initially for cost management?

agrare:
Completely unrelated but I wonder if we could have some helper methods to add the common columns that every entity type table has?

db/migrate/20181130111723_add_volumes.rb
Ladas: We need volume types for AI ops. So for MIQ we've decided to add separate table, because usually it contains more metadata, so name wasn't enough.

db/migrate/20181130111723_add_volumes.rb
Ladas: hm, maybe?

db/migrate/20181130111723_add_volumes.rb
Ladas: btw. these are the types https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html

I think I will also be able to fetch the from the pricing API

db/migrate/20181130111723_add_volumes.rb
Ladas: So do you want to use the cloud_ prefix for the other stuff too? Like networks, interfaces, etc.?

In theory the same volumes table can be used for infra too. Given we'd just have basic set of attributes and the rest is in extra,

db/migrate/20181130111723_add_volumes.rb
agrare: :+1: thanks

db/migrate/20181130111723_add_volumes.rb
agrare: Yeah lets leave just as volume and see if when we get to infra they don't overlap at all we can split it up

db/migrate/20181130111723_add_volumes.rb
```diff
@@ -0,0 +1,55 @@
...
+      t.references :source_region, :type => :bigint, :index => true, :foreign_key => {:on_delete => :cascade}
+
+      t.string :source_ref
+      t.string :name
+      t.string :status
+      t.bigint :size
+      t.jsonb  :extra
+
+      t.timestamps
+      t.datetime   :archived_on
```
agrare: Can you update this to archived_at

Ladas: of course, sorry I forgot I've already merged your PR :-D

agrare:
@Ladas okay this looks great, can we get models and inventory collections for these?

Ladas:
@agrare awesome, I am prototyping refresh for these, then I'll push it

PR #70, opened 2018-11-30
Extract topological_inventory-persister

Fryguy:
cc @agrare

Here is the current extracted repo: https://github.com/ManageIQ/topological_inventory-persister  - It's not ready yet as I need to figure out db rake tasks still.


agrare:
@Fryguy this ready to take out of WIP?

Fryguy:
This can be better as is, and I will make a follow up PR with more specific removals after

Issue #75, opened 2018-12-05
Incorrect use of parameters during provisioning


mkanoor:
https://github.com/ManageIQ/topological_inventory-core/blob/dbceaa98c2499a49a6eb8725d46c1114a2fbe1e5/lib/topological_inventory/core/service_plan_client.rb#L10

The service portal is sending the namespace to use (default for now) for provisioning in the provider control parameters.

The provider control parameters is being merged with service parameters which is incorrect.
https://github.com/eclarizio/topological_inventory-api/blob/2329cb746e79b3950a053b9992cffb775eff10da/app/controllers/api/v0/service_plans_controller.rb#L18

We have to keep the 2 parameters separate till we are ready to order

* service_parameters goes here
    https://github.com/ManageIQ/topological_inventory-core/blob/dbceaa98c2499a49a6eb8725d46c1114a2fbe1e5/lib/topological_inventory/core/service_plan_client.rb#L15

* provider_control_parameters a hash has the 'namespace'
  https://github.com/ManageIQ/topological_inventory-core/blob/dbceaa98c2499a49a6eb8725d46c1114a2fbe1e5/lib/topological_inventory/core/service_plan_client.rb#L10

PR #77, opened 2018-12-06
Rework parameters for sending namespace and service_parameters

eclarizio:
This is to address https://github.com/ManageIQ/topological_inventory-core/issues/75.

It will also need [this API PR](https://github.com/ManageIQ/topological_inventory-api/pull/60)

@mkanoor Let me know if this is what you were thinking.
/cc @agrare

### DIFF
--- a/spec/topological_inventory/core/service_plan_client_spec.rb
+++ b/spec/topological_inventory/core/service_plan_client_spec.rb
@@ -4,18 +4,25 @@ module TopologicalInventory
   module Core
     describe ServicePlanClient do
       describe "#build_payload" do
+        let(:order_parameters) do
+          {
+            :service_parameters => service_parameters,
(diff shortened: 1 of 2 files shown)


spec/topological_inventory/core/service_plan_client_spec.rb
```diff
@@ -4,18 +4,25 @@ module TopologicalInventory
   module Core
     describe ServicePlanClient do
       describe "#build_payload" do
+        let(:order_parameters) do
+          {
+            :service_parameters => service_parameters,
+            :provider_control_parameters => {:namespace => "namespace"}
+          }
+        end
+        let(:service_parameters) { "service_parameters" }
```
mkanoor: @eclarizio This should be a hash, it would help someone looking at the code. The service parameter is specific to a service that you are trying to provision.

PR #79, opened 2018-12-10
Schema for parsed flavor values

Ladas:


Ladas:
cc @agrare , it's a bit simplified comparing to MIQ

PR #83, opened 2019-01-02
Tagging modeling

Ladas:
Tagging modeling without act-as-taggable, using mapping table per entity and foreign_keys with cascade delete.


Ladas:
cc @agrare @Fryguy it's a draft of how this could look like, so it follows the IPP. I've marked several question there as TODO

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20181210164829) do
+ActiveRecord::Schema.define(version: 2019_01_02_161442) do
```
Fryguy: Minor and unrelated, but I would let ActiveRecord do what it does and not listen to rubocop on this one (I thought db/schema.rb was an ignored file anyway)

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
```diff
@@ -0,0 +1,56 @@
...
+  def change
+    create_table "tags", id: :serial, force: :cascade do |t|
+      t.references :tenant, :type => :bigint, :index => false, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string "name", :null => false
+      t.text "description"
+      t.boolean "active", :default => true, :null => false
+
+      # TODO what is human readable name? Will we use it for searching? How will we search with translated field?
+      t.string "display_name"
```
agrare: Not sure if we need to keep the human readable name in our own datastore?  The IPP states this lives with the tagging taxonomy service.  It would be helpful for our own display needs but if it is owned by that service and is changed we would have to keep them in sync.  Maybe for display we would check with the tagging taxonomy service for a display name based on the tag.

db/schema.rb
Ladas: huh, I did't change this manually :-) Seems like new active record generates it like this?

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
Ladas: @Fryguy @agrare Well it's unclear what the display name should do, if I see display name: ```Deployment Environment```, with name: ```env```. I am gonna search for it using the display name, because that is what I see. But then, I'll see a translation, so I'll search using that translation, right?

And if we'll support starts-with, contains, etc. then we can't do some simple mapping back to name.

So any idea how this should work?

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      # TODO the queries like "starts_with" won't be consistent across platform, if we'll model separately vs. encoded
+      # as namespace/name. Or we'll have to do query transformations, will we do that?
+      t.string "namespace"
```
Ladas: @agrare @Fryguy please advise, if we'll model it separately while IPP says it should be prefix, we'l need to transform a lot of queries, e.g. for ```aws/env```:

```starts_with("aws/e") => name.starts_with("aws") && namespace.starts_with("e")```
```contains("/env") => namespace IS NOT NULL && name.contains("env")```

...
so should we rather model it as prefix of the name?

File: db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb [hunk 4]
```diff
@@ -0,0 +1,56 @@
...
+
+      # TODO how to model 'Whether this tag can accept zero, one, or multiple associated values', should be just
+      # 'multiple' boolean and allow it only if there are values to be associated?
+      t.boolean "multiple", :default => false, :null => false
```
Ladas: @Fryguy @agrare please advise. Are we going to add available tag values to it's own table (tag_values)? Or we'll encode the value into the name?

Also will we use the same approach for the mapped tags? So e.g. taking aws tag ```Name``` the tag_values would contain all unique names.

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    create_table "vms_tags", id: :serial, force: :cascade do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.references :tag, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      # TODO should we name the relation according to the table it relates to? Or just always taggable_id?
+      t.references :vm, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
```
Ladas: @agrare @Fryguy please advise. Talking to Adam, we incline to ditch the polymorphic approach, since we can't use foreign keys then. Then the mapping tables should be almost the same, except the reference to specific table ```vm``` in this case. Are we ok with this?

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      # TODO might be better to always have association to value? Where the value could be NULL or empty string? That
+      # way we can make sure the value is always set. Otherwise we could have nil here, if the value is actually required.
+      t.references :tag_value, :type => :bigint, :index => true, :null => true, :foreign_key => {:on_delete => :cascade}
```
Ladas: @agrare @Fryguy please advise, not sure what is better. If omitting the value, in case of no value? Or always having a value, where it would be some default value record with "" or nil.

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
agrare: I'm not sure if we need to support query by human readable name, the IPP doesn't state but I imagined we would only query based on the tag name

Ladas: Ah, so it's more like what you see as a user. If you see tag ```Deployment Environment``` in french. You'll be filtering it using that. But yeah, it just won't work, so we should ditch the display name and make sure users can see the tag name, so they can filter using that

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
agrare: Good point, either way core should handle setting the namespace (either in the tag name or separate) rather than the provider author.

We'll need to split or join the namespace at some point.  If we store it separately we'll need to split what is passed in when doing a query, and if we store it in the tag name we'll need to split it to do any native operations like tagging a vm.

> The key syntax will be a UTF-8 encoded string containing only letters, numbers, hyphens, and underscores. Optionally, a namespace prefix may be attached to the key name and delimited with a slash. 

This sounds like we'd be able to assume that any slash delimits the namespace and the rest of the tag but I'm not sure we can assume that native source tags won't have `'/'` in them.

Ladas: Hm, I've leaned to have it encoded to the name, but your right, we can't ensure provider won't have '/' in the tag name. So it looks like it should be in a separate column after all.

Ladas: Also it looks like the namespace should really be associated with source_type. So it gets assigned automatically. Should we rather associate source or source_type here, instead of namespace?

agrare: Actually we'd probably have an easier time stripping it off the name in the operations worker because we already know the source_type so we could strip exactly e.g. `"aws/"` instead of just trying to match on the first `"/"` in the query

Ladas: Maybe. If you'd call you provider tag as ```aws/env```, we'd store it as ```aws/aws/env```. So it would still work I guess :-)

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
```diff
@@ -0,0 +1,41 @@
+class AddTaggingsForAllTaggableTables < ActiveRecord::Migration[5.1]
+  def change
+    create_table "tags", id: :serial, force: :cascade do |t|
+      t.references :tenant, :type => :bigint, :index => false, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string "name", :null => false
+      t.text "description"
+
+      t.boolean "active", :default => true, :null => false
```
agrare: I wonder if we need to keep active and multiple (similar to dropping display_name)?

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
```diff
@@ -0,0 +1,41 @@
...
+      t.boolean "multiple", :default => false, :null => false
+
+      t.index ["tenant_id", "name"], :unique => true
+    end
+
+    create_table "vms_tags", id: :serial, force: :cascade do |t|
```
agrare: `"vm_tags"` would make all the associations easier, with `VmsTags` we're going to have to define the class name on all the assocs

Ladas:
@agrare if we'll use the polymorphic relation, we can't use foreign keys and cascade delete .So we'd have to code it ourselves (so we can delete in batches) or rely on Rails, but then we'd have to delete per 1 record, which will be slow on tags table (because it will be big).

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
Ladas: you're probably right. I think I'll be able to do our own act_as_a_taggable, what will just generate the right relations names. 

If we'll use ```vm_tags```, we'll need to call singularize on when generationg the inventory collectio name.

app/models/application_record.rb
```diff
@@ -8,7 +7,7 @@ module VERSION
         MAJOR = ActiveRecord::VERSION::MAJOR
       end
     end
-  end unless Rails.const_defined?("VERSION")
+  end if !defined?(::Rails) || !::Rails.const_defined?("VERSION")
```
Ladas: @agrare when I removed the require ^, it was failing on undefined const ```Rails```, the acts-as-taggable was probably requiring rails somewhere

Ladas:
@agrare s othis is ready, basic tagging migrations and model changes for vms, container_groups and container_images

app/models/application_record.rb
agrare: Interesting, this hack was added because the acts_as_tenant gem was using Rails without checking if it was defined first, nice catch.

app/models/application_record.rb
```diff
@@ -8,14 +7,18 @@ module VERSION
...
   require 'acts_as_tenant'
 
   self.abstract_class = true
 
   def as_json(options = {})
     options[:except] ||= []
-    options[:except] << :tag_list
     super
   end
```
agrare: We probably don't need to override this method at all now? cc @bdunne ?

app/models/concerns/act_as_taggable_on.rb
```diff
@@ -0,0 +1,28 @@
+module ActAsTaggableOn
+  def acts_as_taggable_on
+    mapping_table_name = "#{name.underscore}_tags".to_sym
```
agrare: Could we just use `def self.tagging_relation_name` for this?  Any reason for this to be a local variable vs a method?

app/models/application_record.rb
bdunne: Yeah, that was because of `acts_as_taggable`

db/migrate/20190102161442_add_taggings_for_all_taggable_tables.rb
Ladas: we can probably add these later

app/models/concerns/act_as_taggable_on.rb
Ladas: hm I think I can't use the method for the has_many? And I can use the variable in the methods body... Or can I?

agrare: You can you just need to define the method in the `class_eval` before you do the `has_many`.  Would prefer that over having it defined in two places two different ways.

app/models/concerns/act_as_taggable_on.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+
+      def taggings
+        public_send(self.class.tagging_relation_name).map do |tagging|
+          {
+            :tag_id => tagging.tag_id.to_s,
+            :name   => tagging.tag.name,
+            :value  => tagging.value,
+          }
+        end
```
agrare: If we do tags as a subcollection we won't need this, we can figure that out after discussion on the -api PR

app/models/concerns/act_as_taggable_on.rb
Ladas: done

app/models/concerns/act_as_taggable_on.rb
Ladas: ok, removing for now

Ladas:
@agrare so we'll definitely need the table (so the foreign keys work), but we might be able to generate the model as part of the acts as taggable I suppose

agrare:
> so we'll definitely need the table (so the foreign keys work)

Don't want to rule out that there might be other ways of doing this

PR #85, opened 2019-01-09
Log the payload when creating a service instance

carbonin:
cc @gtanzillo this should give us what we're looking for to figure out if we're sending the right parameters when ordering a service

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -14,6 +14,8 @@ require "webmock/rspec"
 require "topological_inventory/core/ar_helper"
 TopologicalInventory::Core::ArHelper.load_environment!
 
+Rails.logger = Logger.new(TopologicalInventory::Core::ArHelper.root.join("log/test.log"))
+
 RSpec.configure do |config|
   config.use_transactional_fixtures = true
   # rspec-expectations config goes here. You can use an alternate
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@
 /config/database.yml
 /coverage/
 /doc/
-/log/
+/log/*.log
 /pkg/
 /spec/reports/
 /tmp/
--- a/lib/topological_inventory/core/ar_helper.rb
+++ b/lib/topological_inventory/core/ar_helper.rb
@@ -31,6 +31,10 @@ module TopologicalInventory
         ActiveRecord::Base.establish_connection(ENV["RAILS_ENV"].to_sym)
       end
 
+      def self.root
+        @root ||= Pathname.new(__dir__).join("../../..").expand_path
+      end
+
       private_class_method def self.load_models
         $LOAD_PATH << root.join("app", "models")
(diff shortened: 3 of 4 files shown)


carbonin:
@bdunne what do you think about making the logger a dev dependency here? I guess the assumption is that a `Rails.logger` will be available wherever this is used in production. Does that make sense?

carbonin:
Also ping @eclarizio

spec/spec_helper.rb
```diff
@@ -14,6 +14,9 @@
 require "topological_inventory/core/ar_helper"
 TopologicalInventory::Core::ArHelper.load_environment!
 
+require 'manageiq/loggers'
+Rails.logger = ManageIQ::Loggers::Base.new(TopologicalInventory::Core::ArHelper.root.join("log/test.log"))
```
bdunne: Wouldn't it make more sense for it to be `TopologicalInventory::Core.root...`?

carbonin: Sure? I just left it where it was defined before. It was a private class method.

bdunne:
> what do you think about making the logger a dev dependency here?

I don't think it's necessary especially since it's just for tests.

carbonin:
>>what do you think about making the logger a dev dependency here?

>I don't think it's necessary especially since it's just for tests.

Okay, but that's what I did. So would you rather me stick it in the `Gemfile` rather than `gemspec`?

bdunne:
>>>what do you think about making the logger a dev dependency here?

>> I don't think it's necessary especially since it's just for tests.

> Okay, but that's what I did. So would you rather me stick it in the Gemfile rather than gemspec?

Do we need `manageiq-loggers` at all?  I don't think so since it's only being used for tests.  I don't think we need the test logs formatted with out formatter.

carbonin:
> Do we need manageiq-loggers at all? I don't think so since it's only being used for tests. I don't think we need the test logs formatted with out formatter.

Good point, I'll make it a regular logger.

eclarizio:
Yeah, seems fine to me to have it be a regular logger since we don't care about the test logs anyway. 👍

spec/spec_helper.rb
```diff
@@ -14,6 +14,8 @@
 require "topological_inventory/core/ar_helper"
 TopologicalInventory::Core::ArHelper.load_environment!
 
+Rails.logger = Logger.new(TopologicalInventory::Core::ArHelper.root.join("log/test.log"))
```
bdunne: Is this necessary?

carbonin: Yes, that's the entire purpose of the PR ... The specs were failing before with no method `#info` for nil.

bdunne: Oh, I didn't know that

carbonin: Ah, sorry, `s/PR/commit/` 

Originally I just added the logging to the method, but then the specs started failing so I had to do this.

PR #86, opened 2019-01-09
Loosen dependencies

bdunne:
- loosen dependency on Rails
- Upgrade and loosen dependency on manageiq-password


topological_inventory-core.gemspec
```diff
@@ -19,9 +19,9 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "acts-as-taggable-on", '~> 6.0'
   s.add_runtime_dependency "acts_as_tenant"
   s.add_runtime_dependency "manageiq-messaging", "~> 0.1.0"
-  s.add_runtime_dependency "manageiq-password",  "~> 0.1.0"
+  s.add_runtime_dependency "manageiq-password",  "~> 0.2"
   s.add_runtime_dependency "pg", "> 0"
-  s.add_runtime_dependency "rails", "~> 5.2.1.1"
```
carbonin: This was probably there to please Hakiri, can we do `> 5.2.1.1` as well?

PR #87, opened 2019-01-11
Upgrade rails to satisfy Hakiri

bdunne:


gtanzillo:
@bdunne Is the expectation that the Hakiri warnings will go away with this change?  Does it need to be run again/?

bdunne:
That was the expectation since it worked on a different repo, but now it shows different problems.

PR #88, opened 2019-01-11
Source type for amazon

Ladas:
Add default source type for amazon and also make source type NOT NULL

Depends on:
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/92

### DIFF
--- a/db/migrate/20190111083026_change_sources_source_type_to_not_null.rb
+++ b/db/migrate/20190111083026_change_sources_source_type_to_not_null.rb
@@ -1,0 +1,5 @@
+class ChangeSourcesSourceTypeToNotNull < ActiveRecord::Migration[5.1]
+  def change
+    change_column_null(:sources, :source_type_id, false)
+  end
+end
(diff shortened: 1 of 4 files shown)


Ladas:
cc @agrare

Ladas:
@agrare should I do data migration for this? (creating the aws source type and filling it everywhere where we have a nil?) Or we'll just do it manually when the migration fails?

agrare:
> should I do data migration for this? (creating the aws source type and filling it everywhere where we have a nil?) Or we'll just do it manually when the migration fails?

I think we should just let it fail since there's no way to know if it was supposed to be AWS or not and both our live deployments have source_type_id present for all sources.

agrare:
@Ladas why does this depend on https://github.com/ManageIQ/topological_inventory-core/pull/92 ?  It doesn't look like it should

Ladas:
@agrare well it all depends on changed schema, when putting it as dependencies I don't need to resolve conflicts for the schema :-)

PR #89, opened 2019-01-11
Don't send template parameters with empty values

carbonin:
Parameters which are not set by the user in the service portal
dialog are being sent to us as empty strings. We should not be sending
those parameters over to OpenShift. This is the way the OpenShift
catalog UI handles the case when a user doesn't fill in a non-required
parameter.

See https://github.com/openshift/origin/issues/21769 for full details of the error case.
cc @eclarizio

### DIFF
--- a/lib/topological_inventory/core/service_plan_client.rb
+++ b/lib/topological_inventory/core/service_plan_client.rb
@@ -1,7 +1,14 @@
+require 'more_core_extensions/core_ext/hash'
+
 module TopologicalInventory
   module Core
     class ServicePlanClient
       def build_payload(service_plan_name, service_offering_name, order_parameters)
+        # We need to not send empty strings in case the parameter is generated
+        # More details are explained in the comment in the OpenShift web catalog
+        # https://github.com/openshift/origin-web-catalog/blob/4c5cb3ee1ae0061ed28fc6190a0f8fff71771122/src/components/order-service/order-service.controller.ts#L442
+        safe_params = order_parameters[:service_parameters].delete_blanks
+
         {
           "apiVersion" => "servicecatalog.k8s.io/v1beta1",
           "kind"       => "ServiceInstance",
@@ -12,7 +19,7 @@ module TopologicalInventory
           "spec"       => {
             "clusterServiceClassExternalName" => service_offering_name,
             "clusterServicePlanExternalName"  => service_plan_name,
-            "parameters"                      => order_parameters[:service_parameters]
+            "parameters"                      => safe_params
           }
         }.to_json
       end
--- a/spec/topological_inventory/core/service_plan_client_spec.rb
+++ b/spec/topological_inventory/core/service_plan_client_spec.rb
@@ -10,7 +10,7 @@ module TopologicalInventory
             :provider_control_parameters => {:namespace => "namespace"}
           }
         end
-        let(:service_parameters) { {"DB_NAME" => "TEST_DB", "namespace" => "TEST_DB_NAMESPACE"} }
+        let(:service_parameters) { {"DB_NAME" => "TEST_DB", "namespace" => "TEST_DB_NAMESPACE", "EMPTY" => ""} }
         let(:expected_payload) do
           {
             "apiVersion" => "servicecatalog.k8s.io/v1beta1",
@@ -22,7 +22,7 @@ module TopologicalInventory
             "spec"       => {
               "clusterServiceClassExternalName" => "service_offering_name",
               "clusterServicePlanExternalName"  => "service_plan_name",
-              "parameters"                      => service_parameters
+              "parameters"                      => {"DB_NAME" => "TEST_DB", "namespace" => "TEST_DB_NAMESPACE"}
             }
           }.to_json
         end
--- a/topological_inventory-core.gemspec
+++ b/topological_inventory-core.gemspec
@@ -20,6 +20,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "acts_as_tenant"
   s.add_runtime_dependency "manageiq-messaging", "~> 0.1.0"
   s.add_runtime_dependency "manageiq-password",  "~> 0.2"
+  s.add_runtime_dependency "more_core_extensions"
   s.add_runtime_dependency "pg", "> 0"
   s.add_runtime_dependency "rails", "~> 5.2", ">= 5.2.1.1"
   s.add_runtime_dependency "rest-client", ">= 1.8.0"


carbonin:
@mkanoor @gmcculloug this should fix the issue we've been seeing when trying to provision templates.

carbonin:
No idea. I think it belongs with whatever bit has the most detailed knowledge of the requirements of this specific payload. But I'm not super attached to this particular solution.

carbonin:
I guess what I'm saying is that I don't want to enforce this generally for service plans across multiple source types when we get to that point. This feels like a very specific corner case for openshift templates so I want to keep this "hack" contained.

gmcculloug:
The values we are passing back are derived from the JSON schema of the dialog parameters.  Since the caller is using Topology as a normalization layer it would seem problematic if the caller had to know intimate details about what different source providers.

carbonin:
Oh, I interpreted this as "should it live in ServiceCatalogClient rather than ServicePlanClient"

bdunne:
Deleting all keys with empty string values seems dangerous and that might not be correct, maybe an empty string is the desired response.  I would assume that a caller to our api would only send the keys that are necessary.

If we're going to modify the incoming request, I think we should at least check it against the JSON Schema so that we aren't deleting any required values.

carbonin:
I think the required-ness is being handled on the caller side already. @gmcculloug ?

eclarizio:
> I guess what I'm saying is that I don't want to enforce this generally for service plans across multiple source types when we get to that point. This feels like a very specific corner case for openshift templates so I want to keep this "hack" contained.

Yeah that should be fine, and I'm going to be moving this into the operations-openshift repo that I'm working on for when this becomes async, so it will be even further "contained" 👍.

bdunne:
@gmcculloug @mkanoor Do we know where the empty strings are coming from?

mkanoor:
@bdunne 
These are optional parameters for a service. In the service plan definition the parameters can be either required or optional. The UI when it displays the Dialog it checks if all the required parameters have been set and sends the required + optional parameters. The optional parameters can be empty strings.  For OpenShift it seems optional empty parameters causes provision to fail for other providers (sources) they might need all the parameters and generate passwords, userid if the field is not filled in.

carbonin:
After speaking with @mkanoor, @gmcculloug, and @bdunne about this, I think I can sum up our options.

1. Service portal sends all parameters to topological inventory with parameters unset by the user as empty strings. This means that whenever an empty string is treated as a value or unset parameters need to be sent as some value other than an empty string, topological inventory will need to write something up to handle the situation (this PR is an example of this option).

2. Service portal only sends default and user-set parameters to topological inventory. In this case a parameter would only be sent with a value of empty string if the user removed a default value from a parameter. This means that whenever a source requires all possible parameters to be sent and unset parameters to be sent as some other value (such as an empty string, nil, 0, etc.) topological inventory will need to write up a special case.

-------------------------------------------------------------------------

After giving these some thought, it seems to me that option 2 is the better way forward.

Sending a parameter with a value of empty string as a place holder for "unset" seems like the exception rather than the rule. It seems much more likely to me that an external system would expect optional parameters to not be sent if a value is not being assigned to them (as OpenShift is doing here).

It seems like the UI would be the only component here able to distinguish the difference between a parameter which had a default value which was removed and a parameter which was unset by the user (and had no default value). So I feel like the best solution would be for the UI to track user changes in the form and only send a parameter if it has a value or was altered by the user.

-------------------------------------------------------------------------

Thoughts @bdunne @mkanoor @gmcculloug @Fryguy and others?

gmcculloug:
Adding @lgalis and @Hyperkid123 for the UI perspective.

mkanoor:
@carbonin 
I think we are making an assumption here that the UI will not enforce required parameters. If a parameter is marked as required the UI will enforce it on the browser side even before the data is sent to the Service Portal.

carbonin:
The required-ness is not relevant @mkanoor The issue is how we are representing an unset non-required value.

mkanoor:
@carbonin That would be a null string.

mkanoor:
Its possible that the Open Service Broker doesn't know how to handle parameters for which the data can be automatically generated. This is the template file for nginx
https://github.com/openshift/library/blob/329466ff9eabc4dcb39eedd22d8f6e3c4885a1d6/official/nginx/templates/nginx-example.json#L266

And that field is an generated field

mkanoor:
@carbonin I tested this with empty fields as long as they are not secret fields it seems to work properly. Secret fields that can be auto generated cannot have empty values. Unfortunately that information seems to be lost when they take a OpenShift template and convert it to a Open Service Broker specification. This is how the OSB service plan looks like, note that the generator is gone
```
- apiVersion: servicecatalog.k8s.io/v1beta1
  kind: ClusterServicePlan
  metadata:
    creationTimestamp: 2018-12-12T09:25:46Z
    name: 80f11d3d-fdee-11e8-860c-06945c5af756
    namespace: ""
    ownerReferences:
    - apiVersion: servicecatalog.k8s.io/v1beta1
      blockOwnerDeletion: false
      controller: true
      kind: ClusterServiceBroker
      name: template-service-broker
      uid: c66c8db0-fdef-11e8-9ddf-0a580a800006
    resourceVersion: "4875"
    selfLink: /apis/servicecatalog.k8s.io/v1beta1/clusterserviceplans/80f11d3d-fdee-11e8-860c-06945c5af756
    uid: e8371ef0-fdef-11e8-9ddf-0a580a800006
  spec:
    bindable: true
    clusterServiceBrokerName: template-service-broker
    clusterServiceClassRef:
      name: 80f11d3d-fdee-11e8-860c-06945c5af756
    description: Default plan
    externalID: 80f11d3d-fdee-11e8-860c-06945c5af756
    externalMetadata:
      schemas:
        service_instance:
          create:
            openshift_form_definition:
            - NAME
            - NAMESPACE
            - NGINX_VERSION
            - MEMORY_LIMIT
            - SOURCE_REPOSITORY_URL
            - SOURCE_REPOSITORY_REF
            - CONTEXT_DIR
            - APPLICATION_DOMAIN
            - GITHUB_WEBHOOK_SECRET
            - GENERIC_WEBHOOK_SECRET
    externalName: default
    free: true
    instanceCreateParameterSchema:
      $schema: http://json-schema.org/draft-04/schema
      additionalProperties: false
      properties:
        APPLICATION_DOMAIN:
          default: ""
          description: The exposed hostname that will route to the nginx service,
            if left blank a value will be defaulted.
          title: Application Hostname
          type: string
        CONTEXT_DIR:
          default: ""
          description: Set this to the relative path to your project if it is not
            in the root of your repository.
          title: Context Directory
          type: string
        GENERIC_WEBHOOK_SECRET:
          default: ""
          description: A secret string used to configure the Generic webhook.
          title: Generic Webhook Secret
          type: string
        GITHUB_WEBHOOK_SECRET:
          default: ""
          description: Github trigger secret.  A difficult to guess string encoded
            as part of the webhook URL.  Not encrypted.
          title: GitHub Webhook Secret
          type: string
        MEMORY_LIMIT:
          default: 512Mi
          description: Maximum amount of memory the container can use.
          title: Memory Limit
          type: string
        NAME:
          default: nginx-example
          description: The name assigned to all of the frontend objects defined in
            this template.
          title: Name
          type: string
        NAMESPACE:
          default: openshift
          description: The OpenShift Namespace where the ImageStream resides.
          title: Namespace
          type: string
        NGINX_VERSION:
          default: "1.12"
          description: Version of NGINX image to be used (1.12 by default).
          title: NGINX Version
          type: string
        SOURCE_REPOSITORY_REF:
          default: ""
          description: Set this to a branch name, tag or other ref of your repository
            if you are not using the default branch.
          title: Git Reference
          type: string
        SOURCE_REPOSITORY_URL:
          default: https://github.com/sclorg/nginx-ex.git
          description: The URL of the repository with your application source code.
          title: Git Repository URL
          type: string
      required:
      - NAME
      - NAMESPACE
      - NGINX_VERSION
      - MEMORY_LIMIT
      - SOURCE_REPOSITORY_URL
      type: object
    serviceBindingCreateParameterSchema:
      $schema: http://json-schema.org/draft-04/schema
      additionalProperties: false
      type: object
  status:
    removedFromBrokerCatalog: false

```

mkanoor:
@lgalis found this in the openshift code which says very clearly that it cannot handle empty strings in generated fields
https://github.com/openshift/origin-web-catalog/blob/4c5cb3ee1ae0061ed28fc6190a0f8fff71771122/src/components/order-service/order-service.controller.ts#L442

bdunne:
>> The required-ness is not relevant @mkanoor The issue is how we are representing an unset non-required value.

> @carbonin That would be a empty string.

Any string (even an empty one) is a value, so it makes sense to me that the generator in Openshift wouldn't override the input that was provided.

carbonin:
Agree with @bdunne, the issue is not in the broker as the generator is run when the template is evaluated. It also isn't limited to secret fields. Any parameter can be set up with a generator.

But these are all openshift specific concerns. My issue is that we need to distinguish between unset and empty-as-a-value. The only way to do this is for the UI to track user choices.

lib/topological_inventory/core/service_plan_client.rb
```diff
@@ -1,7 +1,14 @@
+require 'more_core_extensions/core_ext/hash'
+
 module TopologicalInventory
   module Core
     class ServicePlanClient
       def build_payload(service_plan_name, service_offering_name, order_parameters)
+        # We need to not send empty strings in case the parameter is generated
+        # More details are explained in the comment in the OpenShift web catalog
+        # https://github.com/openshift/origin-web-catalog/blob/4c5cb3ee1ae0061ed28fc6190a0f8fff71771122/src/components/order-service/order-service.controller.ts#L442
+        safe_params = order_parameters[:service_parameters].delete_blanks
```
bdunne: FWIW I still disagree with this approach and think it's going to bite us, but I seem to be out numbered ¯\\\_(ツ)_/¯

mkanoor:
@carbonin can you add reference to this GitHub issue
https://github.com/openshift/origin/issues/14445

carbonin:
@mkanoor I think the linked code and comment is good enough, I don't know that the reasoning behind why it won't be changed really matters to why we're doing it here.

agrare:
Sorry I'm coming in to this late but I'm going to have to agree with @bdunne and @carbonin on this one, an empty string is a value and we must pass it through.

Values that were unset by the user should not be set in the order params IMO.  This is not a source type specific concept, the fact that OpenShift happens to fail is irrelevant (other than to underscore that passing `""` when the user really meant unset is wrong).

carbonin:
I'm going to close this because I think this code is moving to the openshift operations worker, but this is still something that needs to be resolved, right?

gmcculloug:
Yes, it is still needed.

@eclarizio @agrare Can you confirm if this code was applied to the OCP specific repo?

eclarizio:
@gmcculloug I don't think this is currently being handled in the operations worker. The `ServicePlanClient` class and other related classes were moved, but the consideration for empty strings/values hasn't been messed with. Re-reading through this I think I agree that we should be treating anything that is in the `order_params` as "intentionally" set, even if it's an empty string. If the user somehow unsets something, then it shouldn't come through in the order parameters, which I guess means there would need to be some sort of data manipulation before it gets to us, right?

gmcculloug:
@Fryguy @gtanzillo As I recall we beat this issue into the ground multiple times and ended up with the agreement that this was the `least worst option` to address this issue.

Unfortunately we did not capture notes from other discussions here, but is further discussion required at this point?

bdunne:
IIRC, the decision was that we would put the hack (delete blank strings) in the operations code for OpenShift only

gmcculloug:
Correct, I should have clarified that the code would be OpenShift only.  Thanks @bdunne

PR #90, opened 2019-01-11
Source name should be not null

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/88


agrare:
:+1: makes sense, I'll merge when you run the migrations and take out of WIP

PR #92, opened 2019-01-18
Add service offering icons

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/91

### DIFF
--- a/app/models/service_offering.rb
+++ b/app/models/service_offering.rb
@@ -4,7 +4,7 @@ class ServiceOffering < ApplicationRecord
   include ArchivedConcern
 
   belongs_to :tenant
-  belongs_to :service_offering_icon
+  belongs_to :service_offering_icon, :optional => true
   belongs_to :source
   belongs_to :source_region
   belongs_to :subscription
--- a/app/models/service_offering.rb
+++ b/app/models/service_offering.rb
@@ -14,4 +14,8 @@ class ServiceOffering < ApplicationRecord
 
   acts_as_tenant(:tenant)
   acts_as_taggable_on
+
+  def icon_data
+    service_offering_icon&.data
+  end
 end
--- a/app/models/service_offering.rb
+++ b/app/models/service_offering.rb
@@ -4,6 +4,7 @@ class ServiceOffering < ApplicationRecord
   include ArchivedConcern
 
   belongs_to :tenant
+  belongs_to :service_offering_icon
   belongs_to :source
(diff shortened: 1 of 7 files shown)


Ladas:
cc @agrare

app/models/service_offering.rb
```diff
@@ -13,4 +14,8 @@ class ServiceOffering < ApplicationRecord
 
   acts_as_tenant(:tenant)
   acts_as_taggable_on
+
+  def icon_data
+    service_offering_icon&.data
+  end
```
agrare: We're trying to stay away from having methods in the models as much as possible to keep them clean, it is just as easy for the caller to do this.

app/models/service_offering_icon.rb
```diff
@@ -0,0 +1,8 @@
+class ServiceOfferingIcon < ApplicationRecord
+  belongs_to :tenant
```
agrare: Can you add the other side of this association to `Tenant`?

app/models/service_offering.rb
Ladas: right, so this is similar to tags, so faking an attribute, so we can use it in API response

alternative here is to just expose service_offering_icon_id and service_offering_icons collection?

agrare: Ah hm, how about we start with the API matching the modeling then change it if we need to?

Ladas: ok (my thinking was starting with what is most convenient for the user)

db/migrate/20190118110059_add_service_offering_icons.rb
```diff
@@ -0,0 +1,16 @@
+class AddServiceOfferingIcons < ActiveRecord::Migration[5.2]
+  def change
+    create_table "service_offering_icons", :id => :serial, :force => :cascade do |t|
+      t.references :tenant, :type => :bigint, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :type => :bigint, :index => false, :null => false, :foreign_key => {:on_delete => :cascade}
```
agrare: Why not have an index on this one?

Ladas: yeah, adding specs for these. Source is already covered by the unique index

agrare: Ah okay

PR #94, opened 2019-01-22
Test schema health

Ladas:
Depends on:
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/90

### DIFF
--- a/db/migrate/20190124154742_add_missing_not_null_constraints_to_source_ref.rb
+++ b/db/migrate/20190124154742_add_missing_not_null_constraints_to_source_ref.rb
@@ -1,0 +1,21 @@
+class AddMissingNotNullConstraintsToSourceRef < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null :container_images, :source_ref, false
+    change_column_null :orchestration_stacks, :source_ref, false
+    change_column_null :flavors, :source_ref, false
+    change_column_null :vms, :source_ref, false
+    change_column_null :volume_types, :source_ref, false
+    change_column_null :container_groups, :source_ref, false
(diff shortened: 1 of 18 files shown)


agrare:
> Thoughts? Any voices against the gem above? The main motivation is that it's impossible to check these manually (even I am forgetting to add stuff) :-)

This is an awesome idea, haven't looked at the gem yet but I'm all for this.  Only issue I can think of is we might end up having to maintain a huge list of "overrides" for stuff that doesn't conform, especially when we try to merge with the MIQ schema.

Ladas:
@agrare well I would say that if we are going to merge it with MIQ, the specs will have to pass without any overrides. :-) We should just fix the DB.

agrare:
That's optimistic ;)

db/migrate/20190124154742_add_missing_not_null_constraints_to_source_ref.rb
```diff
@@ -0,0 +1,21 @@
+class AddMissingNotNullConstraintsToSourceRef < ActiveRecord::Migration[5.2]
+  def change
+    change_column_null(:container_images, :source_ref, false)
```
agrare: Don't worry about the parens just to appease rubocop.  We just need to add this method to the overrides.

Ladas: hm ok

bdunne: https://github.com/ManageIQ/guides/pull/335

bdunne:
:+1: Sorry for the delay, this looks great

PR #98, opened 2019-02-08
Modeling features availabilities

agrare:
Add a table to track the support and availability of features of resources/actions based on API path.

```
>> openshift_source = Source.create!(:tenant => tenant, :source_type => openshift_source_type, :name => "OCP", :uid => SecureRandom.uuid)
>> openshift_source.availabilities.create!(:action => "GET", :identifier => "/service_instances", :availability => "available", :last_valid_at => Time.now.utc, :last_checked_at => Time.now.utc)
>> openshift_source.availabilities.create!(:action => "GET", :identifier => "/service_offerings", :availability => "available", :last_valid_at => Time.now.utc, :last_checked_at => Time.now.utc)
>> openshift_source.availabilities.create!(:action => "GET", :identifier => "/service_plans", :availability => "available", :last_valid_at => Time.now.utc, :last_checked_at => Time.now.utc)
>> openshift_source.availabilities.create!(:action => "POST", :identifier => "/service_plans#order", :availability => "available", :last_valid_at => Time.now.utc, :last_checked_at => Time.now.utc)
```

Possible values for availability: `["unsupported", "supported", "unavailable", "available"]`

This lets users check if a feature is available in a source-type agnostic way:
```
>> openshift_source.availabilities.find_by(:action => "POST", :identifier => "/service_plans#order").availability == "available"
=> true
```

This also allows the provider to update feature availability if e.g. the servicebroker endpoint isn't enabled on the source.

### DIFF
--- a/app/models/source_type.rb
+++ b/app/models/source_type.rb
@@ -1,3 +1,4 @@
 class SourceType < ApplicationRecord
   has_many :sources
+  has_many :availabilities, :as => :resource, :dependent => :destroy
 end
--- a/app/models/source_type.rb
+++ b/app/models/source_type.rb
@@ -1,4 +1,4 @@
 class SourceType < ApplicationRecord
   has_many :sources
-  has_many :availabilities, :as => :resource, :dependent => :destroy
+  has_many :availabilities, :as => :resource, :dependent => :destroy, :inverse_of => :resource
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -29,6 +29,18 @@ ActiveRecord::Schema.define(version: 2019_02_13_155142) do
     t.index ["tenant_id"], name: "index_authentications_on_tenant_id"
   end
 
+  create_table "availabilities", force: :cascade do |t|
+    t.string "resource_type"
+    t.bigint "resource_id"
+    t.string "action", null: false
+    t.string "identifier", null: false
+    t.string "availability", null: false
+    t.datetime "last_checked_at"
+    t.datetime "last_valid_at"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
   create_table "container_group_tags", id: :serial, force: :cascade do |t|
     t.bigint "tenant_id", null: false
     t.bigint "tag_id", null: false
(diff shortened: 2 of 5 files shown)


agrare:
cc @gtanzillo @Fryguy

app/models/source_type.rb
```diff
@@ -1,3 +1,4 @@
 class SourceType < ApplicationRecord
   has_many :sources
+  has_many :availabilities, :as => :resource, :dependent => :destroy
```
Ladas: @agrare I wonder what the example with source type will be? Using some kind of inheritance? Like:

`if source.source_type.available?(feature) && source.available(feature)?`

Or would we always copy it to source when adding it? The source_type will probably have only `supported/unsupported`

agrare: I was thinking we would copy the source_type availabilities when the source was created

agrare:
@Ladas @gtanzillo rebased

app/models/source_type.rb
Ladas: hm, copying might not be the best as we add more stuff, but lets solve that when we get to it. It won't be hard to migrate to something else

db/schema.rb
```diff
@@ -29,6 +29,19 @@
     t.index ["tenant_id"], name: "index_authentications_on_tenant_id"
   end
 
+  create_table "availabilities", force: :cascade do |t|
+    t.string "resource_type"
+    t.bigint "resource_id"
+    t.string "action", null: false
+    t.string "identifier", null: false
```
Ladas: we should add unique index on the first 4 columns

db/schema.rb
```diff
@@ -29,6 +29,19 @@
     t.index ["tenant_id"], name: "index_authentications_on_tenant_id"
   end
 
+  create_table "availabilities", force: :cascade do |t|
+    t.string "resource_type"
+    t.bigint "resource_id"
```
Ladas: these should be also not null?

agrare: :+1: done

db/schema.rb
agrare: :+1: done

PR #99, opened 2019-02-11
Remove belongs_to :source from tag tables

lpichler:
`belongs_to :source`s were added here https://github.com/ManageIQ/topological_inventory-core/pull/83.

Tables `XXX_tags` are for establishing M:N relations between `XXXs` and `tags` table but they don't need relation to `sources` table.

cc @gtanzillo 

@miq-bot assign @Ladas


Ladas:
:+1: I somehow forgot to remove these

PR #103, opened 2019-02-14
[WIP] Use db triggers for tag cascade deletions

agrare:
Move to a single polymorphic `Tagging` join table and use DB after delete triggers to cleanup associated taggings when a tagged item is deleted.

Uses the hair_trigger gem https://github.com/jenseng/hair_trigger to define the trigger in the mixin that way models can just `include acts_as_taggable_on` and run `rake db:generate_trigger_migration` to get triggers.

This makes no attempt to migrate data out of the join tables or implement a down migration, just for proof of concept.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,9 @@ gemspec
 
 gem "inventory_refresh", :git => "https://github.com/ManageIQ/inventory_refresh", :branch => "master"
 
+# Until https://github.com/ErwinM/acts_as_tenant/pull/192 is released
+gem "acts_as_tenant", :git => "https://github.com/ErwinM/acts_as_tenant", :branch => "master"
+
 #
 # Custom Gemfile modifications
 #
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -1,13 +1,4 @@
 class ApplicationRecord < ActiveRecord::Base
-  # Hack for running acts_as_tenant in a non Rails env.
-  # See https://github.com/ErwinM/acts_as_tenant/pull/192 for proposed fix
-  begin
-    module ::Rails
-      module VERSION
-        MAJOR = ActiveRecord::VERSION::MAJOR
-      end
-    end
-  end if !defined?(::Rails) || !::Rails.const_defined?("VERSION")
   require 'acts_as_tenant'
 
   self.abstract_class = true
--- a/app/models/concerns/act_as_taggable_on.rb
+++ b/app/models/concerns/act_as_taggable_on.rb
@@ -7,7 +7,7 @@ module ActAsTaggableOn
         "taggings".to_sym
       end
 
-      has_many tagging_relation_name, :as => :resource
+      has_many tagging_relation_name, :as => :resource, :inverse_of => :resource
       has_many :tags, :through => tagging_relation_name
 
       def self.taggable?
@@ -15,7 +15,7 @@ module ActAsTaggableOn
       end
 
       trigger.after(:delete) do
-        "DELETE FROM taggings WHERE resource_type='#{self.name}' AND resource_id=OLD.id"
+        "DELETE FROM taggings WHERE resource_type='#{name}' AND resource_id=OLD.id"
       end
     end
   end
--- a/app/models/concerns/act_as_taggable_on.rb
+++ b/app/models/concerns/act_as_taggable_on.rb
@@ -3,19 +3,14 @@ module ActAsTaggableOn
     class_eval do
       require "hair_trigger"
 
-      def self.tagging_relation_name
-        "taggings".to_sym
-      end
-
-      has_many tagging_relation_name, :as => :resource, :inverse_of => :resource
-      has_many :tags, :through => tagging_relation_name
+      has_many :tags, :as => :resource
 
       def self.taggable?
         true
       end
 
       trigger.after(:delete) do
-        "DELETE FROM taggings WHERE resource_type='#{name}' AND resource_id=OLD.id"
+        "DELETE FROM tags WHERE resource_type='#{name}' AND resource_id=OLD.id"
       end
     end
   end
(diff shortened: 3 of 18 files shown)


Gemfile
```diff
@@ -12,6 +12,9 @@ gemspec
 
 gem "inventory_refresh", :git => "https://github.com/ManageIQ/inventory_refresh", :branch => "master"
 
+# Until https://github.com/ErwinM/acts_as_tenant/pull/192 is released
+gem "acts_as_tenant", :git => "https://github.com/ErwinM/acts_as_tenant", :branch => "master"
```
agrare: `hair_trigger` was blowing up trying to access `Rails.application` [here](https://github.com/jenseng/hair_trigger/blob/master/lib/hair_trigger.rb#L25-L27), it had logic to handle the non-Rails case but since we were defining a `Rails` stub it was blowing up.

app/models/application_record.rb
```diff
@@ -1,13 +1,4 @@
 class ApplicationRecord < ActiveRecord::Base
-  # Hack for running acts_as_tenant in a non Rails env.
-  # See https://github.com/ErwinM/acts_as_tenant/pull/192 for proposed fix
-  begin
-    module ::Rails
-      module VERSION
-        MAJOR = ActiveRecord::VERSION::MAJOR
-      end
-    end
-  end if !defined?(::Rails) || !::Rails.const_defined?("VERSION")
```
agrare: See https://github.com/ManageIQ/topological_inventory-core/pull/103/files#r256962930

app/models/concerns/act_as_taggable_on.rb
```diff
@@ -1,25 +1,21 @@
 module ActAsTaggableOn
   def acts_as_taggable_on
     class_eval do
+      require "hair_trigger"
+
       def self.tagging_relation_name
-        "#{name.underscore}_tags".to_sym
+        "taggings".to_sym
```
agrare: Code in topological_inventory-api references the `tagging_relation_name` so this is for compatibility until that is changed, then we can just use a normal `has_many :taggings` down here https://github.com/ManageIQ/topological_inventory-core/pull/103/files#diff-2441abf8e311a53a22e85595c554e4a2R10

agrare:
To test:
```
>> source = Source.create!(:tenant => Tenant.first_or_create!, :source_type => SourceType.find_by(:name => "openshift"), :name => "OCP", :uid => SecureRandom.uuid)
>> container_node = source.container_nodes.create!(:tenant => Tenant.first, :source_ref => "abcd")
>> container_node.taggings.create!(:value => "abcd", :tenant => container_node.tenant, :tag => Tag.create!(:tenant => container_node.tenant, :name => "production"))
=> #<Tagging id: 3, tenant_id: 1, resource_type: "ContainerNode", resource_id: 5, tag_id: 4, value: "abcd", created_at: "2019-02-14 18:52:03", updated_at: "2019-02-14 18:52:03">
>> container_node.tags.pluck(:name)
=> ["production"]
>> Tagging.count
=> 1
>> container_node.delete
=> #<ContainerNode id: 5, source_id: 2, source_ref: "72686b1a-2ea2-4aca-9829-8c024ba37b5c", resource_version: nil, name: nil, cpus: nil, memory: nil, tenant_id: 1, created_at: "2019-02-14 18:51:46", updated_at: "2019-02-14 18:51:46", source_deleted_at: nil, source_created_at: nil, archived_at: nil, resource_timestamp: nil, resource_timestamps: {}, resource_timestamps_max: nil, last_seen_at: nil, lives_on_type: nil, lives_on_id: nil>
>> ContainerNode.count
=> 0
>> Tagging.count
=> 0
```

db/migrate/20190214185856_create_triggers_multiple_tables.rb
```diff
@@ -0,0 +1,105 @@
...
+
+class CreateTriggersMultipleTables < ActiveRecord::Migration[5.2]
+  def up
+    create_trigger("container_groups_after_delete_row_tr", :generated => true, :compatibility => 1)
+      .on("container_groups")
+      .after(:delete) do
+      "DELETE FROM taggings WHERE resource_type='ContainerGroup' AND resource_id=OLD.id;"
+    end
+
+    create_trigger("container_groups_after_delete_row_tr", :generated => true, :compatibility => 1)
```
gtanzillo: Any idea why these are duplicated?

agrare: No idea I'll look into this

agrare:
We also talked about just putting tag names and values in the same table instead of splitting up between tags and taggings.  Since our service isn't where the list of possible tag keys comes from we don't need to maintain a list of tags that aren't associated with any entities so it should work having them all in one table and would be simpler.

app/models/concerns/act_as_taggable_on.rb
```diff
@@ -1,25 +1,21 @@
...
-      def taggings
-        public_send(self.class.tagging_relation_name).map do |tagging|
-          {
-            :tag_id => tagging.tag_id.to_s,
-            :name   => tagging.tag.name,
-            :value  => tagging.value,
-          }
-        end
+      trigger.after(:delete) do
+        "DELETE FROM taggings WHERE resource_type='#{name}' AND resource_id=OLD.id"
```
Ladas: What does this do? We should already have the DB trigger, right?

Ladas:
@agrare so this is like a middle way, we'll keep the cascade delete but we'll lose the foreign key constraint.

Can we sum what are the benefits/drawbacks of this in the main description? 

E.g. for the API, will we expose the whole collection or we'll have a `resource_type` filter mandatory? (we'd could expose it with mandatory filter with a current way).

This will reduce the number of needed models and table definitions. But I suppose we can generate a lot of code. And we'll have one huge table, vs. several smaller.

Ladas:
@lpichler if we'll go this way, this will probably affect what you do with tagging

app/models/concerns/act_as_taggable_on.rb
agrare: Not if we go to a single taggings table or collapse taggings and tags into just tags.

agrare:
Closing in favor of https://github.com/ManageIQ/topological_inventory-core/pull/104

agrare:
Closing in favor of moving value to tags and making it a polymorphic assoc

PR #104, opened 2019-02-18
[WIP] Move value from intermediate tables to tags

agrare:
Update the intermediate tables (e.g. container_group_tags, vm_tags) to be simple join table, moving the value property to the tags table.

### DIFF
--- a/app/models/concerns/act_as_taggable_on.rb
+++ b/app/models/concerns/act_as_taggable_on.rb
@@ -11,16 +11,6 @@ module ActAsTaggableOn
       def self.taggable?
         true
       end
-
-      def taggings
-        public_send(self.class.tagging_relation_name).map do |tagging|
-          {
-            :tag_id => tagging.tag_id.to_s,
-            :name   => tagging.tag.name,
-            :value  => tagging.value,
-          }
-        end
-      end
     end
   end
 end
--- a/db/migrate/20190218180413_move_tag_value_to_tags.rb
+++ b/db/migrate/20190218180413_move_tag_value_to_tags.rb
@@ -1,0 +1,40 @@
+class MoveTagValueToTags < ActiveRecord::Migration[5.2]
+  def up
+    add_column :tags, :value, :string, :null => false, :default => ""
+
+    remove_column :container_group_tags,    :value
+    remove_column :container_image_tags,    :value
+    remove_column :container_node_tags,     :value
+    remove_column :container_project_tags,  :value
+    remove_column :container_template_tags, :value
+    remove_column :service_offering_tags,   :value
+    remove_column :vm_tags,                 :value
+  end
+
+  def down
+    remove_column :tags, :value
+
+    add_column :container_group_tags,    :value, :string, :default => "", :null => false
+    add_column :container_image_tags,    :value, :string, :default => "", :null => false
+    add_column :container_node_tags,     :value, :string, :default => "", :null => false
+    add_column :container_project_tags,  :value, :string, :default => "", :null => false
+    add_column :container_template_tags, :value, :string, :default => "", :null => false
+    add_column :service_offering_tags,   :value, :string, :default => "", :null => false
+    add_column :vm_tags,                 :value, :string, :default => "", :null => false
+
+    add_index :container_group_tags, [:container_group_id, :tag_id, :value],
+              :name => "uniq_index_on_container_group_id_tag_id_and_value", :unique => true
+    add_index :container_image_tags, [:container_image_id, :tag_id, :value],
+              :name => "uniq_index_on_container_image_id_tag_id_and_value", :unique => true
+    add_index :container_node_tags, [:container_node_id, :tag_id, :value],
+              :name => "uniq_index_on_container_node_id_tag_id_and_value", :unique => true
+    add_index :container_project_tags, [:container_project_id, :tag_id, :value],
+              :name => "uniq_index_on_container_project_id_tag_id_and_value", :unique => true
+    add_index :container_template_tags, [:container_template_id, :tag_id, :value],
+              :name => "uniq_index_on_container_template_id_tag_id_and_value", :unique => true
+    add_index :service_offering_tags, [:service_offering_id, :tag_id, :value],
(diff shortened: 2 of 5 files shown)


db/migrate/20190218180413_move_tag_value_to_tags.rb
```diff
@@ -0,0 +1,40 @@
+class MoveTagValueToTags < ActiveRecord::Migration[5.2]
+  def up
+    add_column :tags, :value, :string, :null => false, :default => ""
```
Ladas: we'll also need to change the unique index, now it's 
`["tenant_id", "namespace", "name"]`
we'll need
`["tenant_id", "namespace", "name", "value"]`

(you might need to drop&recreate it)

db/migrate/20190218203848_remove_extra_columns_from_tagging_join_tables.rb
```diff
@@ -0,0 +1,19 @@
...
+  def change
+    remove_column :container_group_tags, :created_at, :datetime
+    remove_column :container_image_tags, :created_at, :datetime
+    remove_column :container_node_tags, :created_at, :datetime
+    remove_column :container_project_tags, :created_at, :datetime
+    remove_column :container_template_tags, :created_at, :datetime
+    remove_column :service_offering_tags, :created_at, :datetime
+    remove_column :vm_tags, :created_at, :datetime
+
+    remove_reference :container_group_tags, :tenant, :index => true, :null => false,  :foreign_key => {:on_delete => :cascade}
```
Ladas: @kbrock @Fryguy what is the requirement for the `tenant_id` should it be in every table? Were we planning to do partitioning/sharding based on this key?

db/migrate/20190218180413_move_tag_value_to_tags.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    remove_column :container_group_tags,    :value
```
Ladas: and here we're just removing the index (not sure if this line removes it), but we need to add the new unique index that will be on `["container_group_id", "tag_id"]` and same for others
`

db/migrate/20190218180413_move_tag_value_to_tags.rb
agrare: :+1:

db/migrate/20190218180413_move_tag_value_to_tags.rb
agrare: Yeah it looks like there was only an index on the individual columns, probably should be on the pair of columns.

db/migrate/20190218203848_remove_extra_columns_from_tagging_join_tables.rb
agrare: If it is just a join table and never exposed I don't think we need a tenant

db/migrate/20190218180413_move_tag_value_to_tags.rb
Ladas: the indexed looked like 
```t.index ["container_node_id", "tag_id", "value"], name: "uniq_index_on_container_node_id_tag_id_and_value", unique: true```

the remove column is probably removing it

Comment (agrare) [reply db/migrate/20190218180413_move_tag_value_to_tags.rb]: Interesting okay there was a separate one then just on tag_id then also, `t.index ["tag_id"], name: "index_container_group_tags_on_tag_id"`

File: db/schema.rb [hunk 4]
```diff
@@ -509,7 +479,8 @@
     t.string "namespace", default: "", null: false
     t.text "description"
     t.datetime "created_at", null: false
-    t.index ["tenant_id", "namespace", "name"], name: "index_tags_on_tenant_id_and_namespace_and_name", unique: true
+    t.string "value", default: "", null: false
```
Ladas: how did this got here? :-) I don't see it in a migration though

agrare:
On Wed, Feb 20, 2019 at 4:04 PM, Ladislav Smola 
<notifications@github.com> wrote:
> @Ladas commented on this pull request.
> 
> In db/schema.rb:
> 
> > @@ -509,7 +479,8 @@
>      t.string "namespace", default: "", null: false
>      t.text "description"
>      t.datetime "created_at", null: false
> -    t.index ["tenant_id", "namespace", "name"], name: 
> "index_tags_on_tenant_id_and_namespace_and_name", unique: true
> +    t.string "value", default: "", null: false
> how did this got here? :-) I don't see it in a migration though
> 

https://github.com/ManageIQ/topological_inventory-core/pull/104/files#diff-59899171a8d39115abcf95a8a5b362cbR3

agrare:
@Ladas we'll need to update the inventory collections here and the collectors

kbrock:
Right now, tags table is kinda useless. It is just the materialized full name from the classification. Could we merge this gook into classification and drop tag all together?

kbrock:
ok - sorry. I thought this was working with miq tags - it is working with vm tags.
disregard my comment. As far as I can see, 👍 good stuff 👍

app/models/concerns/act_as_taggable_on.rb
```diff
@@ -11,16 +11,6 @@ def self.tagging_relation_name
       def self.taggable?
         true
       end
-
-      def taggings
```
Ladas: we should keep this to keep 0.0 API working? And we'll need to modify it

db/schema.rb
```diff
@@ -43,14 +43,9 @@
   end
 
   create_table "container_group_tags", id: :serial, force: :cascade do |t|
```
Ladas: we will need also unique indexes here, let me add that in my PR

db/migrate/20190218203848_remove_extra_columns_from_tagging_join_tables.rb
Ladas: hm, removing the tenant_id is not revertable, since it's NOT_NULL

agrare:
Closing in favor of https://github.com/ManageIQ/topological_inventory-core/pull/109

PR #119, opened 2019-03-14
Fix issues with api clients looking for Rails.logger

agrare:
The API clients initial configuration looks to see if Rails constant is defined and if it is will use Rails.logger.  Since we hack Rails::VERSION but do not set Rails.logger this causes any OpenAPI clients to throw an exception on initial configuration:

```
	4: from bin/sources-sync:33:in `<main>'
	3: from /home/agrare/.gem/bundler/gems/sources-api-client-ruby-632713f8cc89/lib/sources-api-client.rb:48:in `configure'
	2: from /home/agrare/.gem/bundler/gems/sources-api-client-ruby-632713f8cc89/lib/sources-api-client/configuration.rb:153:in `default'
	1: from /home/agrare/.gem/bundler/gems/sources-api-client-ruby-632713f8cc89/lib/sources-api-client/configuration.rb:153:in `new'
/home/agrare/.gem/bundler/gems/sources-api-client-ruby-632713f8cc89/lib/sources-api-client/configuration.rb:146:in `initialize': undefined method `logger' for Rails:Module (NoMethodError)
```

### DIFF
--- a/lib/topological_inventory/core/logging.rb
+++ b/lib/topological_inventory/core/logging.rb
@@ -1,0 +1,19 @@
+require "manageiq/loggers"
+
+module TopologicalInventory
+  module Core
+    class << self
+      attr_writer :logger
+    end
+
+    def self.logger
+      @logger ||= ManageIQ::Loggers::Container.new
+    end
+
+    module Logging
+      def logger
+        TopologicalInventory::Core.logger
+      end
+    end
+  end
+end
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -6,6 +6,10 @@ class ApplicationRecord < ActiveRecord::Base
       module VERSION
         MAJOR = ActiveRecord::VERSION::MAJOR
       end
+      def self.logger
+        require "topological_inventory/core/logger"
+        TopologicalInventory::Core.logger
+      end
     end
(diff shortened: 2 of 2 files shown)


lib/topological_inventory/core/logging.rb
```diff
@@ -0,0 +1,19 @@
+require "manageiq/loggers"
+
+module TopologicalInventory
+  module Core
+    class << self
+      attr_writer :logger
+    end
+
+    def self.logger
+      @logger ||= ManageIQ::Loggers::Container.new
```
Fryguy: @agrare I recently put a PR into a different repo dealing with logger in test mode.  I wonder if it's better for this line to just be 

```ruby
@logger ||= Rails.env.test? ? Logger.new("/dev/null") : ManageIQ::Loggers::Container.new
```

Fryguy: Additionally, I wonder if we should do this in the logger helper from manageiq-api-common, otherwise we have to duplicate this setup all over the place.

See https://github.com/ManageIQ/manageiq-api-common/blob/master/lib/manageiq/api/common/logging.rb and https://github.com/ManageIQ/topological_inventory-api/blob/master/config/application.rb#L45

PR #121, opened 2019-03-14
Delete models which were moved to Sources service

agrare:
Update topological_inventory-core models for the extracted sources service: https://github.com/agrare/sources-api


db/migrate/20190314153151_extract_sources_to_new_service.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    drop_table :applications
+    drop_table :application_types
+    drop_table :authentications
+    drop_table :availabilities
+    drop_table :endpoints
+    drop_table :source_types
+  end
+
+  def down
```
Fryguy: Needs the down still as discussed.

agrare: Done

PR #124, opened 2019-03-25
external_tenant must be not null and unique

Ladas:
Fixing specs in Persister: https://github.com/ManageIQ/topological_inventory-persister/pull/21


agrare:
@Ladas looks like you just need to update the specs

Ladas:
@agrare ah right

agrare:
This is where FactoryBot would come in handy

PR #125, opened 2019-03-29
Add "role" to source_type seeds.

martinpovolny:
The `role` value is needed when creating `Endpoint` instances. Until recently this has beed hard-wired in the client.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -10,6 +10,7 @@ end
 openshift_json_schema = {
   :title  => "Configure OpenShift",
   :fields => [
+    {:component => "text-field", :name => "role", :type => "hidden", :initialValue => "kubernetes"},
     {:component => "text-field", :name => "url", :label => "URL"},
     {:component => "checkbox", :name => "verify_ssl", :label => "Verify SSL"},
     {:component => "text-field", :name => "certificate_authority", :label => "Certificate Authority", :condition => {:when => "verify_ssl", :is => true}},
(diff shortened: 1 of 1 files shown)


db/seeds.rb
```diff
@@ -10,6 +10,7 @@ def update_or_create(model, attributes)
 openshift_json_schema = {
   :title  => "Configure OpenShift",
   :fields => [
+    {:component => "text-field", :name => "role", :type => "hidden", initialValue: "kubernetes"},
```
agrare: Should be `:initialValue => "kubernetes"`

agrare:
@martinpovolny the rubocop wanings look legit, otherwise LGTM

martinpovolny:
@agrare : I was following a style in that file (not started by me ;-)).

I take it as you want that fixed (in the whole file then).

agrare:
> @agrare : I was following a style in that file (not started by me ;-)).
>
> I take it as you want that fixed (in the whole file then).

Interesting, where else was it like that in that file I don't see it?

martinpovolny:
> Interesting, where else was it like that in that file I don't see it?

It was completely elsewhere. Sorry.

martinpovolny:
Here it was: 
https://github.com/martinpovolny/manageiq-api-common/blob/6ec9b673ac5709a0e26a49779bf3e9b1534bd7ce/spec/lib/manageiq/api/common/paginated_response_spec.rb#L36

My memory failed me.

PR #130, opened 2019-04-16
Source's settings for storing collector settings

slemrmartin:
MockSource's settings can be customized through YAML in string.
- [ ] https://github.com/ManageIQ/topological_inventory-mock_source/pull/17/

These custom source's settings have to be stored somewhere for orchestrator.

### DIFF
--- a/db/migrate/20190416123117_add_config_to_endpoint.rb
+++ b/db/migrate/20190416123117_add_config_to_endpoint.rb
@@ -1,5 +1,0 @@
-class AddConfigToEndpoint < ActiveRecord::Migration[5.2]
-  def change
-    add_column :endpoints, :config, :string
-  end
-end
--- a/db/migrate/20190416123117_add_config_to_endpoint.rb
+++ b/db/migrate/20190416123117_add_config_to_endpoint.rb
@@ -1,0 +1,5 @@
+class AddConfigToEndpoint < ActiveRecord::Migration[5.2]
+  def change
+    add_column :endpoints, :config, :string
+  end
+end
--- a/db/migrate/20190416123117_add_settings_to_source.rb
+++ b/db/migrate/20190416123117_add_settings_to_source.rb
@@ -1,0 +1,5 @@
+class AddSettingsToSource < ActiveRecord::Migration[5.2]
+  def change
+    add_column :sources, :settings, :jsonb
+  end
+end


slemrmartin:
cc @agrare @carbonin

carbonin:
Agree with @agrare here. And more generally if we're adding this to core I would expect this to apply to (or at least potentially be useful for) all collectors.

If the only use case for this is the mock collector, then I would argue that it doesn't belong in the schema at all.

slemrmartin:
@agrare I'm open to both, sources and endpoint. I chose Endpoint because it seems to me there are settings for Source - Sources UI are using it - and it seems to be 1:1 relationship now - and orchestrator creates ENDPOINT_... ENV variables, which will fit my requirements.

About jsonb - I think that jsonb is not ok for this purpose - config is YAML, it's converted to string either from ConfigMap or other way. So then there has to be several yaml-string-jsonb-string-yaml conversions

agrare:
@slemrmartin Just so I understand, your high-level goal is for the orchestrator to be able to start mock collectors?

slemrmartin:
yes, I create simple rails app, which will create many mock sources through API. Then orchestrator starts collectors. 
Or is there better way?

slemrmartin:
...I mean many mock sources with custom settings

slemrmartin:
> Agree with @agrare here. And more generally if we're adding this to core I would expect this to apply to (or at least potentially be useful for) all collectors.
> 
> If the only use case for this is the mock collector, then I would argue that it doesn't belong in the schema at all.
We need to instantiate for example 1000 sources (mock) with custom settings. So I need to store it somewhere. 
`Source.settings` as jsonb will be ok? Then I'll need some custom code in orchestrator to convert json to yaml to string

agrare:
@slemrmartin lets have a meeting to discuss more generally

slemrmartin:
@carbonin  @agrare After I saw ConfigMap example from Nick, I don't think it's much useful for MockSource. 
First step remains the same: Send custom config through API  and save in db (i.e. in Souces.settings jsonb - this PR). 
Then orchestrator has to either define configmap and extends deployment config:
- it's less universal for all collectors than ENV var
- all config values have to be defined

or stores JSONB to ENV variable:
- ENV variable seems to be more universal 
- for MockSource it can contain only changes against default config files
- but: MockSource expects string with YAML (but I think `Yaml.load` parses Json string too)

PR #131, opened 2019-05-14
Add Host and Cluster models

agrare:
Add initial tables for hosts and clusters, needed to add vsphere and ovirt inventory.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -49,11 +49,11 @@ ActiveRecord::Schema.define(version: 2019_05_14_165841) do
     t.datetime "resource_timestamps_max"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.datetime "archived_on"
+    t.datetime "archived_at"
     t.datetime "source_created_at"
     t.datetime "source_deleted_at"
     t.datetime "last_seen_at"
-    t.index ["archived_on"], name: "index_clusters_on_archived_on"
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -28,6 +28,38 @@
...
+    t.string "source_ref", null: false
+    t.string "uid_ems"
+    t.string "name"
+    t.jsonb "extra"
+    t.datetime "resource_timestamp"
+    t.jsonb "resource_timestamps", default: {}
+    t.datetime "resource_timestamps_max"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.datetime "archived_on"
```
Ladas: @agrare this is `archived_at` right? (maybe we could add this to the schema health specs, making sure there are no `_on` timestamps)

db/schema.rb
```diff
@@ -277,6 +309,45 @@
...
+    t.string "power_state"
+    t.bigint "cpus"
+    t.bigint "memory"
+    t.jsonb "extra"
+    t.datetime "resource_timestamp"
+    t.jsonb "resource_timestamps", default: {}
+    t.datetime "resource_timestamps_max"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.datetime "archived_on"
```
Ladas: same archived_at

db/schema.rb
agrare: Ah right I forgot we fixed that, fixed

PR #140, opened 2019-07-24
Add network_adapters table and models

agrare:
Add a table for network adapters so that we can keep track of NIC mac
addresses and ipaddresses


agrare:
@Ladas I forget are we able to handle polymorphic tables now?  If not I'll have to create identical host_network_adapters and vm_network_adapters tables

PR #141, opened 2019-07-30
Add more networking models

Ladas:
Changing network adapters and adding more networking models, namely floating ips, security groups, subnets and networks

Depends on:
- [x] https://github.com/ManageIQ/topological_inventory-core/pull/143


db/migrate/20190729194551_change_network_adapters.rb
```diff
@@ -0,0 +1,54 @@
+class ChangeNetworkAdapters < ActiveRecord::Migration[5.2]
+  def change
+    add_reference :network_adapters, :source, :index => false, :null => false, :foreign_key => {:on_delete => :cascade}
+    add_reference :network_adapters, :orchestration_stack, :index => true, :null => true, :foreign_key => {:on_delete => :cascade}
+
+    # Relations to VMs, Hosts, etc.
+    add_reference :network_adapters, :device, :index => true, :polymorphic => true, :null => true
```
Ladas: @agrare so I have the other approach commented below, but for now I think I prefer the polymorphic approach. Let me try how that looks with the persister schema. The other approach would need a lot of unique indexes, if we'd add mode devices. Also the relation is just `{:on_delete => :nullify}` so it's not a big deal that it's polymorphic

agrare: Yeah I agree the polymorphic assoc seems to be the best if we can make it work

db/migrate/20190729214242_add_ipaddresses_subnets_and_networks.rb
```diff
@@ -0,0 +1,150 @@
...
+      t.datetime :source_deleted_at
+      t.datetime :last_seen_at
+
+      t.index :archived_at
+      t.index :last_seen_at
+
+      t.index %i[source_id source_ref], :unique => true
+    end
+
+    create_table :ipaddresses, id: :bigserial, force: :cascade do |t|
```
Ladas: @agrare shouldn't it be `ip_addresses` ? OR do you like `ipaddresses` more?

agrare: I think `ipaddresses` is fine

agrare:
@Ladas do we need different `Ipaddress` and `FloatingIp` models?

Ladas:
@agrare I think we do, floating ip as an assignable object living on it's own and IP address as a per VM thing. 

Although the network_adapters are kind_of the same thing now and we model them in 1 table.

I'll try to first finish refresh with separate floating_ip, then I can see how it looks like if I merge them together. The ipaddresses would then need source_id and tenant_id cols.

Ladas:
@agrare hm, damn trying to put together ipaddresses and floating_ips, I think we'll have issue with also the network_adapters

Example:
given we have indexes like:
```
 add_index :network_adapters,
              %i(source_id source_ref device_id device_type),
              :unique => true,
              :name   => "unique_index_network_adapters_with_device_id_and_type"
    add_index :network_adapters,
              %i(source_id source_ref),
              :where  => "device_id IS NULL AND device_type IS NULL",
              :unique => true,
              :name   => "unique_index_network_adapters"
```

Now if we'll have in AWS a network adapter `NetworkAdapter.create(:source_id => 1, :source_ref => "adapter_1_source_ref")`, which would be a isolated network adapter. 

Now if we assign the adapter, it will actually cause `NetworkAdapter.create(:source_id => 1, :source_ref => "adapter_1_source_ref", :device_id => 1, :device_type => "Vm")` and it will archive the old adapter. That is because finding the adapter by `source_id, source_ref and device` will return []

So that is probably not what we want. we want the same NetworkAdapter to be used, for AWS. For infra, we'd want it being archived and recreated.

Solutions
------------------------------ 

1. Just using `:source_id, :source_ref` and unique index and for infra case, building them to be unique under source. While storing the original ids into extra, so we can easily use them for operations or other API calls.

2. Somehow exposing what :manager_ref to use for building the records. So for AWS network_adapters I could say just use [:source_ref, :source_id] for lookup/uniq index. (and e.g. for ipaddresses, some build objects would use `[:source_id, :source_ref]` and some would use `[:source_id, :source_ref, :network_adapter_id, :subnet_id]`) (this would probably require to store this on inventoryCollection level, so we'd need several InventoryCollection variations)

------------

I am thinking that 2. looks cleaner, but harder to understand. So maybe we should just build the artificial :source_ref, but I know you don't like it :-) (but maybe with storing the original attrs in extra, it's ok? What other concerns did you have?)

PR #142, opened 2019-08-02
Add datastores to persister default schema

agrare:


Ladas:
@agrare the schema contains also DatastorMount and DatastoreTag, do you have also those?

PR #146, opened 2019-08-07
Add Source#collector_status string for use in TPINVTRY-532

bdunne:


db/migrate/20190807183701_add_source_collector_status.rb
```diff
@@ -0,0 +1,5 @@
+class AddSourceCollectorStatus < ActiveRecord::Migration[5.2]
+  def change
+    add_column :sources, :collector_status, :string
```
agrare: @bdunne I wonder if we can cover a few cases here, how about instead of collector_status we had a `refresh_status` and `last_refresh_date`?
That way we can handle the "collector failed to deploy" case as well as the "refresh failed because of xyz" or "refresh was successfully completed at aa:bb`

bdunne:
@agrare I renamed the column as suggested and added a validation including the values that I am planning to use.  I did not add a `last_refresh_timestamp` column since I'm not going to populate that with the orchestrator.

PR #147, opened 2019-08-08
Prevent re-requiring models leading to warnings in test

bdunne:
Previously...
```
$ rake
app/models/source.rb:67: warning: already initialized constant Source::ALLOWED_REFRESH_STATUS_VALUES
app/models/source.rb:67: warning: previous definition of ALLOWED_REFRESH_STATUS_VALUES was here
```

### DIFF
--- a/lib/topological_inventory/core/ar_helper.rb
+++ b/lib/topological_inventory/core/ar_helper.rb
@@ -40,7 +40,7 @@ module TopologicalInventory
         $LOAD_PATH << root.join("app", "models", "concerns")
 
         ActiveSupport::Dependencies.autoload_paths << root.join("app", "models")
-        ActiveSupport::Dependencies.autoload_paths << root.join("app", "models", "concenrs")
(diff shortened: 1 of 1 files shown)


agrare:
@bdunne This was added so that non-rails apps that used -core (e.g. the persister) didn't have to require all the models since there wasn't any autoloading.  Is that no longer an issue?

bdunne:
@agrare Autoload works after calling `load_environment!` 
```
$ irb -Ilib
irb(main):001:0> require 'topological_inventory/core/ar_helper.rb'; TopologicalInventory::Core::ArHelper.load_environment!
=> #<ActiveRecord::ConnectionAdapters::ConnectionPool:0x0000000002fe3060 @mon_owner=nil, ...>
irb(main):002:0> Source
=> Source (call 'Source.connection' to establish a connection)
irb(main):003:0> Container
=> Container (call 'Container.connection' to establish a connection)
```

agrare:
> Autoload works after calling load_environment!

Why are you yelling? haha

That's awesome

PR #151, opened 2019-09-24
Fix the Source.datastore_mounts through relation

agrare:
The datastore_mounts relation was defined through :hosts before the
hosts relation was defined.


Ladas:
@agrare you'll need to kick the build for persister and API?

agrare:
@Ladas :+1: done

PR #152, opened 2019-09-26
Add service inventories

Ladas:


db/migrate/20190926144242_add_service_inventories_table.rb
```diff
@@ -0,0 +1,30 @@
+class AddServiceInventoriesTable < ActiveRecord::Migration[5.2]
+  def change
+    create_table :service_inventories, :id => :bigserial, :force => :cascade do |t|
+      t.references :tenant, :index => true, :null => false, :foreign_key => {:on_delete => :cascade}
+      t.references :source, :index => false, :null => false, :foreign_key => {:on_delete => :cascade}
+
+      t.string :source_ref, :null => false
+      t.string :name
+      t.string :inventory_type
```
agrare: What are some examples of inventory_type?  Wondering if this should be in extra

agrare:
@Ladas I think you're missing the `ServiceInventory` model

agrare:
We should probably have `service_inventory_tags`

Ladas:
@Ladas ah yeah, the model escaped the git add

Not sure about tags, there are none in tower. Is catalog planning to store tags here?

db/migrate/20190926144242_add_service_inventories_table.rb
Ladas: hm, maybe? In the account I have it's all `inventory`, so not sure what the types are

agrare:
Yeah we wouldn't collect tags from the source, but the user might still want to assign tags to an inventory.  I've heard assigning the "production" tag to an inventory to invoke different checks.

db/migrate/20190926144242_add_service_inventories_table.rb
Ladas: I say this would distinct e.g. AWS inventory, inventory as a script, etc.

agrare: I'm not sure what that means :)

Ladas: ah ok, so it's basically binary now, let me move it to extra

`Type: Identifies whether it is a standard inventory or a Smart Inventory.`

https://docs.ansible.com/ansible-tower/latest/html/userguide/inventories.html#smart-inventories

agrare: :+1:

Ladas: done

PR #153, opened 2019-10-01
Add service offering and instance nodes

Ladas:
The nodes model graphs of offerings and instances and
all of these relate to service inventories.

Depends on:
- [ ] https://github.com/ManageIQ/inventory_refresh/pull/87


slemrmartin:
@Ladas `ServiceInventory` doesn't have defined all `has_many` associations

Ladas:
@slemrmartin yeah, I've been defining just those that will be used in API. With the child_nodes that I find out I can't use in the API, due to way we build queries in the API (we don't use the actual relations)

slemrmartin:
@Ladas I think that both sides of associations should be defined in all models, regardless if you're using them just now or not. It helps understanding of db/models structure to others. I'll create PR for that.

Ladas:
@slemrmartin I don't think it's necessary, but if you'll insist, then we should have spec checking schema vs. models. It's dull to check this manually. :-)

PR #158, opened 2019-11-08
Tables for service credentials

slemrmartin:
TPINVTRY-744

Credentials for ServiceOffering, ServiceOfferingNode, ServiceInstance, ServiceInstanceNode


db/migrate/20191108135108_add_service_credential_tables.rb
```diff
@@ -0,0 +1,58 @@
...
+  def change
+    create_table :service_credential_types, :id => :bigserial, :force => :cascade do |t|
+      t.references :tenant, :index => true, :null => false, :foreign_key => { :on_delete => :cascade }
+      t.references :source, :index => true, :null => false, :foreign_key => { :on_delete => :cascade }
+
+      t.string :source_ref, :null => false
+      t.string :name
+      t.text :description
+
+      t.jsonb :extra # For credentials input fields
```
slemrmartin: not sure about this, AnsibleTower has several booleans for this, but don't know about other provider types (if applicable)

db/migrate/20191108135108_add_service_credential_tables.rb
```diff
@@ -0,0 +1,58 @@
...
+      t.references :tenant, :index => true, :null => false, :foreign_key => { :on_delete => :cascade }
+      t.references :source, :index => true, :null => false, :foreign_key => { :on_delete => :cascade }
+
+      t.references :service_credential_type, :index => true, :null => true, :foreign_key => { :on_delete => :nullify }
+
+      t.string :source_ref, :null => false
+      t.string :name
+      t.text :description
+
+      t.jsonb :permissions
```
slemrmartin: not sure about this, AnsibleTower has several booleans for this, but don't know about other provider types (if applicable)

agrare: Hey @slemrmartin could you post an example of what ansible_tower has?  Should help us figure out what the modeling should look like.

@gmcculloug is catalog going to use the permissions for anything?

db/migrate/20191108135108_add_service_credential_tables.rb
```diff
@@ -0,0 +1,58 @@
+class AddServiceCredentialTables < ActiveRecord::Migration[5.2]
+  def change
+    create_table :service_credential_types, :id => :bigserial, :force => :cascade do |t|
```
agrare: What are the different credential types?  Is this something that needs to be a separate model or can it just be a string column on the credential table

db/migrate/20191108135108_add_service_credential_tables.rb
gmcculloug: I have no use-cases for credentials right now other than showing what credentials were used when a job ran.  We have not talked about `permissions`.

agrare: Okay lets only save what we need for now @slemrmartin , keep it simple

slemrmartin: I was thinking that it'll be nice to have more info in UI than "Martin's credentials" (for example as tooltip or modal)

db/migrate/20191108135108_add_service_credential_tables.rb
slemrmartin: The important thing is the "extra" column, which contains form input's description of credentials - needed for ordering the service with "prompt on launch" on credentials

db/migrate/20191108135108_add_service_credential_tables.rb
slemrmartin: @agrare https://18.188.178.206/api/v2/credentials/
```
                "user_capabilities": {
                    "edit": true,
                    "delete": true,
                    "copy": true,
                    "use": true
                },
```

slemrmartin: or maybe "object_roles"? I'm not sure, documentation says nothing about that, so better to remove, right :)

slemrmartin: @gmcculloug one use-case is ordering service through catalog. Some Jobs/Workflows needs user input for credentials in the same way as for inventory and survey. So this is needed if we want fully functional Catalog ordering

gmcculloug: Correct, but Catalog is not supporting Templates from tower that have input requirements for credential and/or inventory for GA.  Likely a future requirement but I do not have a Catalog use case for it at the moment.

slemrmartin: okay, as this is not a big amount of work, we can be prepared for future requirements

db/migrate/20191108135108_add_service_credential_tables.rb
slemrmartin: @agrare these input types are not used for ordering, so I removed credential types at all as you wrote

PR #164, opened 2020-01-17
Update README to reflect organization move to RedHatInsights

gtanzillo:


slemrmartin:
hi @gtanzillo, the CodeClimate URLs are not working, it seems that it's still under ManageIQ org there.

PR #165, opened 2020-01-22
Add reference to tenant in the tag join classes

bdunne:
Previously querying the join table based on tag attributes wasn't working when tenancy was enabled.  The query was incorrectly using a different id as the tenant_id in a select.

Specifically see $1 in this line:
```sql
SELECT COUNT(*) FROM "container_node_tags" WHERE ((SELECT "tags"."namespace" FROM "tags" WHERE "tags"."tenant_id" = $1 AND "tags"."id" = "container_node_tags"."tag_id") = 'test') AND "container_node_tags"."container_node_id" = $1  [["container_node_id", 1174]]
```

From:
```
    [----] D, [2020-01-21T16:56:01.518858 #163322:11a4608] DEBUG -- :   Tenant Load (0.2ms)  SELECT  "tenants".* FROM "tenants" WHERE "tenants"."external_tenant" = $1 LIMIT $2  [["external_tenant", "177"], ["LIMIT", 1]]
    [----] D, [2020-01-21T16:56:01.520881 #163322:11a4608] DEBUG -- :   ContainerNode Load (0.2ms)  SELECT  "container_nodes".* FROM "container_nodes" WHERE "container_nodes"."tenant_id" = $1 AND "container_nodes"."id" = $2 LIMIT $3  [["tenant_id", 21806], ["id", 1174], ["LIMIT", 1]]
    [----] D, [2020-01-21T16:56:01.523646 #163322:11a4608] DEBUG -- :    (0.4ms)  SELECT COUNT(*) FROM "container_node_tags" WHERE ((SELECT "tags"."namespace" FROM "tags" WHERE "tags"."tenant_id" = $1 AND "tags"."id" = "container_node_tags"."tag_id") = 'test') AND "container_node_tags"."container_node_id" = $1  [["container_node_id", 1174]]
    [----] D, [2020-01-21T16:56:01.524642 #163322:11a4608] DEBUG -- :   ContainerNodeTag Load (0.4ms)  SELECT  "container_node_tags".* FROM "container_node_tags" WHERE ((SELECT "tags"."namespace" FROM "tags" WHERE "tags"."tenant_id" = $1 AND "tags"."id" = "container_node_tags"."tag_id") = 'test') AND "container_node_tags"."container_node_id" = $1 ORDER BY "container_node_tags"."id" ASC LIMIT $2 OFFSET $3  [["container_node_id", 1174], ["LIMIT", 100], ["OFFSET", 0]]
```

### DIFF
--- a/app/models/cluster_tag.rb
+++ b/app/models/cluster_tag.rb
@@ -1,4 +1,6 @@
 class ClusterTag < ApplicationRecord
   belongs_to :cluster
   belongs_to :tag
+  belongs_to :tenant
+  acts_as_tenant(:tenant)
 end
--- a/app/models/concerns/act_as_taggable_on.rb
+++ b/app/models/concerns/act_as_taggable_on.rb
@@ -61,6 +61,10 @@ module ActAsTaggableOn
         self.class.tagging_relation_name.to_s.classify.constantize
       end
     end
+
+    tenant_klass = ActsAsTenant.tenant_klass
+    join_klass = tagging_relation_name.to_s.singularize.classify.safe_constantize
+    join_klass.before_validation(:on => :create) { |i| i.send("#{tenant_klass}=", i.tag.send(tenant_klass)) }
   end
 
   def tagged_with(tag_name, options = {})
--- a/db/migrate/20200122181811_add_tenant_to_join_tables.rb
+++ b/db/migrate/20200122181811_add_tenant_to_join_tables.rb
@@ -1,0 +1,134 @@
+class AddTenantToJoinTables < ActiveRecord::Migration[5.2]
+  class Tag < ActiveRecord::Base
+  end
+
+  class ClusterTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class ContainerGroupTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class ContainerImageTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class ContainerNodeTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class ContainerProjectTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class ContainerTemplateTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class DatastoreTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class HostTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class IpaddressTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class NetworkAdapterTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class NetworkTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class ReservationTag < ActiveRecord::Base
+    belongs_to :tag
+  end
+
+  class SecurityGroupTag < ActiveRecord::Base
+    belongs_to :tag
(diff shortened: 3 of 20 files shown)


app/models/concerns/act_as_taggable_on.rb
```diff
@@ -61,6 +61,9 @@ def model_tag_class
         self.class.tagging_relation_name.to_s.classify.constantize
       end
     end
+
+    join_klass = tagging_relation_name.to_s.singularize.classify.safe_constantize
+    join_klass.before_validation { |i| i.tenant = i.tag.tenant }
```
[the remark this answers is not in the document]
bdunne: Updated to use the configured name from ActsAsTenant

bdunne:
> So this looks fine to me as the join tables technically do map to a tenant, so including the id in the join table doesn't really cause a problem, but I'm still curious why this happened as it feels like it should work without this change.
> 
> Can you link to the code that exposed this issue?

It looks like the gem doesn't account for `has_many xxx, :through => yyy` relations.  https://github.com/ErwinM/acts_as_tenant/blob/master/lib/acts_as_tenant/model_extensions.rb#L120-L123 looks like it's the issue.

carbonin:
@bdunne okay, but what were we doing that caused this issue? Can we reproduce it in a general case and maybe fix it in the gem? (separately from this patch)

carbonin:
@agrare want to take a look at this? Or maybe @Fryguy ? Want at least one other opinion on this one.

bdunne:
It seems to be related to all of the AREL in the filtering logic.  I wasn't able to reproduce the problem with simple `where` conditions in the gem.

carbonin:
:man_facepalming: Okay, so do we think the better solution for this is reverting to using `where` over arel in the filtering or doing this?

I would prefer using `where` whenever possible. I didn't understand the need for arel in the filtering code when @bdunne showed it to me. Maybe @abellotti can explain that a bit?

bdunne:
I'd prefer the normal `where` for readability too, but I don't know what problems we were trying to solve by switching to AREL.

db/migrate/20200122181811_add_tenant_to_join_tables.rb
```diff
@@ -0,0 +1,134 @@
...
+      NetworkAdapterTag,
+      NetworkTag,
+      ReservationTag,
+      SecurityGroupTag,
+      ServiceInventoryTag,
+      ServiceOfferingTag,
+      SubnetTag,
+      VmTag,
+    ].each do |klass|
+      klass.find_each { |i| i.update(:tenant_id => i.tag.tenant_id) }
```
lpichler:
```ruby

...=> i.tag...

```

 it looks like that this is doing N+1 queries but maybe it is ok at this moment.

carbonin:
@agrare any thoughts here? @bdunne are we settled that this is probably the way forward to deal with the mysterious issue of tenancy + tagging?

bdunne:
> @agrare any thoughts here? @bdunne are we settled that this is probably the way forward to deal with the mysterious issue of tenancy + tagging?

Yes, I think this is the best path forward.

app/models/cluster_tag.rb
```diff
@@ -1,4 +1,6 @@
 class ClusterTag < ApplicationRecord
   belongs_to :cluster
   belongs_to :tag
+  belongs_to :tenant
```
agrare: Could we add the `belongs_to` to the `act_as_taggable_on` mixin ?

bdunne: Implemented in https://github.com/RedHatInsights/topological_inventory-core/pull/167

PR #167, opened 2020-02-12
Deduplicate tag join relation declarations

bdunne:
The only downside is that the relations don't exist until the other class is loaded because we have eagerloading disabled.

```ruby
$ rails c
Loading development environment (Rails 5.2.4.1)
irb(main):001:0> pp ClusterTag.reflect_on_all_associations
[]
=> []

irb(main):002:0> Cluster
=> Cluster (call 'Cluster.connection' to establish a connection)

irb(main):003:0> pp ClusterTag.reflect_on_all_associations; nil
[#<ActiveRecord::Reflection::BelongsToReflection:0x0000000005c31630
  @active_record=
   ClusterTag (call 'ClusterTag.connection' to establish a connection),
  @association_scope_cache=
   #<Concurrent::Map:0x0000000005c31388 entries=0 default_proc=nil>,
  @constructable=true,
  @foreign_key="cluster_id",
  @foreign_type=nil,
  @klass=nil,
  @name=:cluster,
  @options={},
  @plural_name="clusters",
  @scope=nil,
  @type=nil>,
 #<ActiveRecord::Reflection::BelongsToReflection:0x0000000005c2f268
  @active_record=
   ClusterTag (call 'ClusterTag.connection' to establish a connection),
  @association_scope_cache=
   #<Concurrent::Map:0x0000000005c2f0d8 entries=0 default_proc=nil>,
  @constructable=true,
  @foreign_key="tag_id",
  @foreign_type=nil,
  @klass=nil,
  @name=:tag,
  @options={},
  @plural_name="tags",
  @scope=nil,
  @type=nil>,
 #<ActiveRecord::Reflection::BelongsToReflection:0x0000000005c25c90
  @active_record=
   ClusterTag (call 'ClusterTag.connection' to establish a connection),
  @association_scope_cache=
   #<Concurrent::Map:0x0000000005c25b78 entries=0 default_proc=nil>,
  @constructable=true,
  @foreign_type=nil,
  @klass=nil,
  @name=:tenant,
  @options={},
  @plural_name="tenants",
  @scope=nil,
  @type=nil>]
=> nil
```

### DIFF
--- a/app/models/cluster_tag.rb
+++ b/app/models/cluster_tag.rb
@@ -1,6 +1,2 @@
 class ClusterTag < ApplicationRecord
-  belongs_to :cluster
-  belongs_to :tag
-  belongs_to :tenant
-  acts_as_tenant(:tenant)
 end
--- a/app/models/concerns/act_as_taggable_on.rb
+++ b/app/models/concerns/act_as_taggable_on.rb
@@ -65,6 +65,10 @@ module ActAsTaggableOn
     tenant_klass = ActsAsTenant.tenant_klass
     join_klass = tagging_relation_name.to_s.singularize.classify.safe_constantize
     join_klass.before_validation(:on => :create) { |i| i.send("#{tenant_klass}=", i.tag.send(tenant_klass)) }
+    join_klass.belongs_to(name.underscore.to_sym)
+    join_klass.belongs_to(:tag)
+    join_klass.belongs_to(tenant_klass)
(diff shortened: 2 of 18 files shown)


lpichler:
Is this good practice to put things(like `belongs_to :tenant, ... `) which are used usually on model (or in mixin) to outside of its definition ? Despite fact that it looks neat but also it is hidden  - Maybe preferable way could be <Mixin> and `include <Mixin>` and also I am curious to see others opinion.

PR #182, opened 2020-03-13
Removed extraneous indexes

slemrmartin:
Fixes spec
https://github.com/RedHatInsights/topological_inventory-core/blob/master/spec/topological_inventory/core/database_health_spec.rb#L22

Removed indexes to credentials (similar to tag join tables indexes)


slemrmartin:
cc @lpichler

lpichler:
thanks 👍

agrare:
nice @slemrmartin

PR #183, opened 2020-03-13
Refresh state relations

slemrmartin:
TPINVTRY-837

---

Items are collected in `sweep scopes` (for mark & sweep algorithm) and each model's data in payload are recorded to `refresh_state_parts` table. 

It means that saved items can be paired with specific payload.

### DIFF
--- a/lib/topological_inventory/schema/base.rb
+++ b/lib/topological_inventory/schema/base.rb
@@ -4,6 +4,11 @@ require "inventory_refresh/persister"
 module TopologicalInventory
   module Schema
     class Base < InventoryRefresh::Persister
+      def persist!
+        link_data_to_refresh_state_part
+        super
+      end
+
       def to_hash
         {
           "schema": {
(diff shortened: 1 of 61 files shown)


slemrmartin:
Based on #182.

@agrare I added only "last refresh state" and not a "last successful refresh state", because refresh_state_part ID can be added to inventory objects **before** saving. 

Save successful refresh state means resaving all data again which would slow down the whole process

agrare:
Looks good @slemrmartin I don't see a last_refresh_state change though did you forget to push something or is that a followup?

slemrmartin:
ah, right, I forget to push `inventory_refresh` PR

slemrmartin:
@agrare this can be merged without inventory refresh's PR. 
I'm not sure if `"inventory_refresh", "~> 0.3.2"` in gemspec is enough, or it should be updated.

I meanif it's possible that builder of image for openshift could use cache and doesn't update dependency..?

slemrmartin:
@agrare my comments ^^ was wrong. Persister changes must be implemented to core's persister.

So it's ready

slemrmartin:
* depends on https://github.com/RedHatInsights/topological_inventory-core/pull/184

lib/topological_inventory/schema/base.rb
```diff
@@ -21,6 +26,20 @@ def self.klass_for(name)
         # checks whether it loads what we want
         klass if klass.to_s == klass_name
       end
+
+      private
+
+      def link_data_to_refresh_state_part
```
agrare: @slemrmartin I wonder if this makes more sense in the topological_inventory-persister?  Thinking maybe in this [method](https://github.com/RedHatInsights/topological_inventory-persister/blob/master/lib/topological_inventory/persister/workflow.rb#L87-L98)

PR #186, opened 2020-03-17
Refresh Tracking Timestamps

slemrmartin:
**Issue**: https://github.com/RedHatInsights/topological_inventory-persister/issues/52

Adding two timestamps to `RefreshState` model:
- **started_at** - Time of the beginning of collecting one data set [collector part]
- **finished_at** - Time after saving all parts of one data set [persister part]

Note: Linking data to refresh_state_part moved to persister (https://github.com/RedHatInsights/topological_inventory-persister/pull/53)

---

- [x] **depends on** https://github.com/ManageIQ/inventory_refresh/pull/91

### DIFF
--- a/lib/topological_inventory/schema/base.rb
+++ b/lib/topological_inventory/schema/base.rb
@@ -4,14 +4,6 @@ require "inventory_refresh/persister"
 module TopologicalInventory
   module Schema
     class Base < InventoryRefresh::Persister
-      attr_accessor :refresh_time_tracking
-
-      def initialize(manager)
-        self.refresh_time_tracking = {:persister_started_at => Time.now.utc.to_datetime.to_s}
-        super(manager)
-      end
-
-      # TODO: add time_tracking?
       def to_hash
         {
           "schema": {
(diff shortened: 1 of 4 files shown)


lib/topological_inventory/schema/base.rb
```diff
@@ -18,26 +21,28 @@ def to_hash
...
-        klass_name = "TopologicalInventory::Schema::#{name}"
-        klass = klass_name.safe_constantize
-
-        # sometimes autoloader is confused and loads class with another namespace
-        # checks whether it loads what we want
-        klass if klass.to_s == klass_name
-      end
+      class << self
+        def from_hash(persister_data, manager)
+          persister = super(persister_data, manager)
```
agrare: If you're passing the same arguments you don't need to explicitly pass them to `super`

lib/topological_inventory/schema/base.rb
```diff
@@ -18,26 +21,28 @@ def to_hash
...
+          %i[
+            refresh_state_part_collected_at
+            refresh_state_part_sent_at
+            refresh_state_started_at
+            refresh_state_sent_at
+            ingress_api_sent_at
+          ].each do |timestamp_name|
+            persister.refresh_time_tracking[timestamp_name] = persister_data[timestamp_name.to_s]
```
agrare: This feels like it belongs in the inventory_refresh gem [here](https://github.com/ManageIQ/inventory_refresh/blob/master/lib/inventory_refresh/persister.rb#L140-L145)

slemrmartin: I also tought about it, but this tracking seems to be topological-inventory oriented...but :+1:

agrare: Yeah that version of inventory_refresh is all topo specific already, and when we bring that back to manageiq it will be the whole topo system (persisters and ingress-api)

PR #187, opened 2020-04-13
Updating Task in ServiceInstance custom_save_block

slemrmartin:
**Issue**: https://github.com/RedHatInsights/topological_inventory-ansible_tower/issues/83

Updates _running_ `Tasks` of associated `ServiceInstances`.  
Feature is moved from topological_inventory-ansible_tower operations worker because there isn't timeout.

`Task` extended by `source_id`, `target_source_ref` and `target_type`(`== ServiceInstance`)

Uses custom_save_block for `InventoryCollection :service_instances`.

**Note**: This PR processes Task's status in ansible-tower's specific way. Solved in #188 

---

- [x] **depends on** https://github.com/RedHatInsights/topological_inventory-core/issues/190 **[must be applied to stable branch first]**
- [ ] **depends on** https://github.com/ManageIQ/inventory_refresh/pull/92 **[must be released to RubyGems]**
---

**Gem version _1.1.0_** 
- [ ] released to **RubyGems**

### DIFF
--- a/db/migrate/20200414123737_extract_tasks_source_ref_from_context.rb
+++ b/db/migrate/20200414123737_extract_tasks_source_ref_from_context.rb
@@ -1,0 +1,21 @@
+class ExtractTasksSourceRefFromContext < ActiveRecord::Migration[5.2]
+  def up
+    Task.find_in_batches.each do |tasks|
+      tasks.each do |task|
+        source_ref = task.context&.dig('service_instance', 'source_ref')
+        next if source_ref.nil?
+
+        say "Task: #{task.id}, Source ref: #{source_ref}"
+        source_id = task.context&.dig('service_instance', 'source_id')
+
+        task.update_attributes(:target_source_ref => source_ref.to_s,
+                               :target_type       => 'ServiceInstance',
+                               :source_id         => source_id)
+      end
+    end
+  end
+
+  def down
+    Task.update_all(:source_id => nil, :target_type => nil, :target_source_ref => nil)
+  end
+end
--- a/db/migrate/20200414123737_extract_tasks_source_ref_from_context.rb
+++ b/db/migrate/20200414123737_extract_tasks_source_ref_from_context.rb
@@ -1,17 +1,15 @@
 class ExtractTasksSourceRefFromContext < ActiveRecord::Migration[5.2]
   def up
-    Task.find_in_batches.each do |tasks|
-      tasks.each do |task|
-        source_ref = task.context&.dig('service_instance', 'source_ref')
-        next if source_ref.nil?
+    Task.find_each do |task|
+      source_ref = task.context&.dig('service_instance', 'source_ref')
+      next if source_ref.nil?
 
-        say "Task: #{task.id}, Source ref: #{source_ref}"
-        source_id = task.context&.dig('service_instance', 'source_id')
+      say "Task: #{task.id}, Source ref: #{source_ref}"
+      source_id = task.context&.dig('service_instance', 'source_id')
 
-        task.update_attributes(:target_source_ref => source_ref.to_s,
-                               :target_type       => 'ServiceInstance',
-                               :source_id         => source_id)
-      end
+      task.update_attributes(:target_source_ref => source_ref.to_s,
+                             :target_type       => 'ServiceInstance',
+                             :source_id         => source_id)
     end
   end
 
--- a/lib/topological_inventory/schema/default.rb
+++ b/lib/topological_inventory/schema/default.rb
@@ -1,5 +1,5 @@
 require "topological_inventory/schema/base"
-require "pry-byebug"
+
 module TopologicalInventory
   module Schema
     class Default < TopologicalInventory::Schema::Base
@@ -136,9 +136,8 @@ module TopologicalInventory
             InventoryRefresh::SaveCollection::Base.send(:save_inventory, inventory_collection)
 
             # Get running tasks
-            # TODO: add B-tree index to "context" or make 'source_id' as a separate column
-            tasks_source_ref = Task.where(:state => 'running', :target_type => 'ServiceInstance').where("context->'service_instance'->>'source_id' = ?", source.id.to_s)
-                                 .pluck(Arel.sql("context->'service_instance'->>'source_ref'")) #target_source_ref)
+            tasks_source_ref = Task.where(:state => 'running', :target_type => 'ServiceInstance', :source_id => source.id)
+                                 .pluck(:target_source_ref)
 
             # Load saved service instances (IDs needed)
             svc_instances_values = ServiceInstance.where(:source_ref => tasks_source_ref).pluck(:id, :external_url, :source_ref, Arel.sql("extra->'finished'"), Arel.sql("extra->'status'"))
@@ -152,11 +151,9 @@ module TopologicalInventory
               state = finished_timestamp.blank? ? 'running' : 'completed'
               status = %w[error failed].include?(status) ? 'error' : 'ok' # TODO: ansible-tower specific, normalize in collector
               context = {
-                :remote_status => status,
+                :remote_status    => status,
                 :service_instance => {
-                  :id => id,
-                  :source_id => source.id,
-                  :source_ref => source_ref,
+                  :id  => id,
                   :url => external_url
                 }
               }.to_json
--- a/lib/topological_inventory/schema/default.rb
+++ b/lib/topological_inventory/schema/default.rb
@@ -171,8 +171,7 @@ module TopologicalInventory
               FROM (VALUES :values
               ) AS c(source_ref, state, status, context)
               WHERE t.target_source_ref = c.source_ref
-                AND t.target_type = 'ServiceInstance'
-                AND t.state = 'running';
+                AND t.target_type = 'ServiceInstance';
 SQL
             sql.sub!(':values', sql_update_values.join(','))
 
--- a/lib/topological_inventory/schema/default.rb
+++ b/lib/topological_inventory/schema/default.rb
@@ -1,5 +1,5 @@
 require "topological_inventory/schema/base"
-
+require "pry-byebug"
 module TopologicalInventory
   module Schema
     class Default < TopologicalInventory::Schema::Base
@@ -20,7 +20,7 @@ module TopologicalInventory
         add_default_collection(:networks)
         add_default_collection(:orchestration_stacks)
         add_default_collection(:reservations)
-        add_default_collection(:service_instances)
+        add_service_instances
         add_default_collection(:service_instance_nodes)
         add_default_collection(:service_inventories)
         add_default_collection(:security_groups)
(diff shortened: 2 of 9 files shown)


lib/topological_inventory/schema/default.rb
```diff
@@ -123,6 +123,63 @@ def add_tagging_collection(model, manager_ref: [:source_ref])
...
+
+          save_block = lambda do |source, inventory_collection|
+            src_refs = inventory_collection.data.collect { |inventory_object| inventory_object.source_ref }
+            return if src_refs.blank?
+
+            # Saving Service Instances as usual
+            InventoryRefresh::SaveCollection::Base.send(:save_inventory, inventory_collection)
+
+            # Get running tasks
+            tasks_source_ref = Task.where(:state => 'running', :target_type => 'ServiceInstance', :source_id => source.id)
```
slemrmartin, line 139: @agrare is it worth adding index to `:target_type` if there is `[:target_type, :target_source_ref]`? I don't know which index is used by default (https://www.postgresql.org/docs/10/indexes-multicolumn.html) but it seems all of them can be used for subset of columns.

lib/topological_inventory/schema/default.rb
```diff
@@ -123,6 +123,63 @@ def add_tagging_collection(model, manager_ref: [:source_ref])
...
+        add_collection(:service_instances) do |builder|
+          add_default_properties(builder)
+          add_default_values(builder)
+
+          save_block = lambda do |source, inventory_collection|
+            src_refs = inventory_collection.data.collect { |inventory_object| inventory_object.source_ref }
+            return if src_refs.blank?
+
+            # Saving Service Instances as usual
+            InventoryRefresh::SaveCollection::Base.send(:save_inventory, inventory_collection)
```
agrare, line 136: @slemrmartin I think a better way to do this would be to have another collection e.g. `service_instance_tasks` and define service_instances as a dependent collection

lib/topological_inventory/schema/default.rb
```diff
@@ -123,6 +123,63 @@ def add_tagging_collection(model, manager_ref: [:source_ref])
...
+UPDATE tasks AS t SET
+  state = c.state,
+  status = c.status,
+  context = c.context
+FROM (VALUES :values
+) AS c(source_ref, state, status, context)
+WHERE t.target_source_ref = c.source_ref
+  AND t.target_type = 'ServiceInstance'
+  AND t.state = 'running';
+SQL
```
agrare, line 173: Yeah this isn't a good idea :)

slemrmartin: why not? raw sql update is much faster than using active record - and core is mostly about speed

agrare: Because it is better to be easy to understand and maintain than to be fast 95% of the time.  If it turns out that this block of code is taking up a significant amount of time during a refresh we can optimize it then.  There is also a lot of optimizations that can be done within activerecord without dropping into pure sql.

slemrmartin: afaik there'll be similar upsert in Rails 6 (but not only update). Rails 5 doesn't have bulk update.

db/migrate/20200414123737_extract_tasks_source_ref_from_context.rb
```diff
@@ -0,0 +1,21 @@
+class ExtractTasksSourceRefFromContext < ActiveRecord::Migration[5.2]
+  def up
+    Task.find_in_batches.each do |tasks|
```
agrare, line 3: Super minor, but if you're going to do `.find_in_batches { |tasks| tasks.each { |task| } }` this is basically just `.find_each { |task| }`

lib/topological_inventory/schema/default.rb
slemrmartin: @agrare is this refactoring correct for now as we discussed?

lib/topological_inventory/schema/default.rb
```diff
@@ -123,6 +124,76 @@ def add_tagging_collection(model, manager_ref: [:source_ref])
...
+            service_instance_tasks_custom_save(source, tasks_collection)
+          end
+
+          builder.add_properties(:custom_save_block => save_block)
+        end
+      end
+
+      def service_instance_tasks_custom_save(source, tasks_collection)
+        service_instance_collection = tasks_collection.dependency_attributes[:service_instances]&.first
+        src_refs = service_instance_collection.data.collect { |inventory_object| inventory_object.source_ref }
```
agrare, line 146: `&.first` [here](https://github.com/RedHatInsights/topological_inventory-core/pull/187/files#diff-bf581c67a3bfa3b364180f998a222a83R145) implies this might be nil, but then you do service_instance_collection.data.collect which would fail

lib/topological_inventory/schema/default.rb
agrare: I thought we agreed to go the active_record route and leave the equivalent sql commented out

slemrmartin: ok, I wasn't sure

lib/topological_inventory/schema/default.rb
```diff
@@ -123,6 +124,102 @@ def add_tagging_collection(model, manager_ref: [:source_ref])
...
+      # This method is updating one by one using ActiveRecord
+      def service_instance_tasks_update_ineffective(service_instances_values, tasks_id, tasks_source_ref)
+        tasks = {}
+        service_instances_values.each do |attrs|
+          id, external_url, source_ref, finished_timestamp, status = attrs[0], attrs[1], attrs[2], attrs[3], attrs[4]
+
+          task_id = tasks_id[tasks_source_ref.index(source_ref)]
+          tasks[task_id] = task_update_values(id, external_url, status, finished_timestamp)
+
+          Task.update(tasks.keys, tasks.values)
```
agrare, line 190: I assume you didn't mean for this to be in the loop

lib/topological_inventory/schema/default.rb
```diff
@@ -123,6 +124,102 @@ def add_tagging_collection(model, manager_ref: [:source_ref])
...
+        end
```
agrare, line 191: I think this is more ineffective than it has to be:

```ruby
# Set up 1000 running tasks and service instances
1_000.times { source.service_instances.create!(:tenant => t, :source => s, :source_ref => SecureRandom.uuid) }
source.service_instances.each { |si| Task.create!(:state => "running", :status => "ok", :target_type => "ServiceInstance", :source => s, :target_source_ref => si.source_ref, :tenant => t) }

tasks_by_source_ref = Task.where(:state => 'running', :target_type => 'ServiceInstance', :source_id => source.id, :target_source_ref => service_instances.pluck(:source_ref)).index_by(&:target_source_ref)
service_instances.find_in_batches { |group| ActiveRecord::Base.transaction { group.each { |service_instance| tasks_by_source_ref[service_instance.source_ref].update(:context => {:service_instance => {:id => ser
vice_instance.id}}) } } }
```

Takes `0.70144`

The way you have here:
```
service_instances_values.each do |attrs|
  id, external_url, source_ref, finished_timestamp, status = attrs[0], attrs[1], attrs[2], attrs[3], attrs[4]
  task_id = tasks_id[tasks_source_ref.index(source_ref)]
  tasks[task_id] = {:context => {:service_instance => {:id => id}}}
  end
Task.update(tasks.keys, tasks.values)
```

Takes `4.05301` seconds

slemrmartin: yeah, my mistake with the update in the loop...

ok, so let's waste time with benchmarking:

**Your way**
`@real=1.7165850750170648, @cstime=0.0, @cutime=0.0, @stime=0.08524299999999996, @utime=1.2653870000000005, @total=1.3506300000000004`

**Current way**
`@real=6.16914131003432, @cstime=0.0, @cutime=0.0, @stime=0.28006200000000003, @utime=2.8168819999999997, @total=3.0969439999999997`

**Current way with transaction**
`@real=2.2340267669642344, @cstime=0.0, @cutime=0.0, @stime=0.10396899999999976, @utime=1.7059960000000007, @total=1.8099650000000005`

**Raw sql**
`@real=0.031836505979299545, @cstime=0.0, @cutime=0.0, @stime=0.0009470000000000312, @utime=0.0013389999999979807, @total=0.002285999999998012`

- raw SQL query is **much much better** than any other solution
- current solution encapsulated to transaction is not as better as yours, but not significantly

If you still don't want raw SQL, can I just use this one with transaction? I don't want to spent another time with rewriting, testing and solving conflicts with rebasing in 2nd task

agrare: > I don't want to spend another time with rewriting, testing and solving conflicts with rebasing in 2nd task

If you don't have time for this I'll take over this PR for you

I'd rather not just wrap what you have in a transaction and call it done because the whole point is to make this simple and easy to read, what you have here isn't simple and easy to read

slemrmartin: okay, I don't get it, we discussed that it's not sure that raw sql is faster without benchmarking. Now benchmarking proved it's more than 50x faster and still it isn't good :D 
In your code, it will increase persister's IC saving by 0.5+ seconds... :thinking: 

This PR is not logical for me...

But OK, I'll rewrite it again

agrare: I think you're missing the point a bit, no one ever said that raw sql wasn't faster but that is a bit like taking a formula 1 car to go to the store.

1000 running tasks in each batch of 1000 service instances isn't realistic, but like we've said all along if it turns out that this is significantly increasing the refresh performance overall then we can optimize it.

Do everything the simple and obvious way first, then measure and optimize what needs to be optimized.

While this is slower it might account for 0.01% of the total refresh time overall (for all collections not just this one).  If it ends up accounting for more like 10% then absolutely we should optimize it.

lib/topological_inventory/schema/default.rb
```diff
@@ -123,6 +124,106 @@ def add_tagging_collection(model, manager_ref: [:source_ref])
...
+              :id => svc_instance_id,
+              :url => external_url
+            }
+          }
+        }
+      end
+
+      # This method is updating one by one using ActiveRecord
+      def service_instance_tasks_update_ineffective(source, svc_instances_source_ref)
+        service_instances = ServiceInstance.where(:source_id => source.id, :source_ref => svc_instances_source_ref)
```
agrare, line 170: Since this is a dependent collection on service_instances we should have the ids here from the inventory collection but we can look at that in a followup

lib/topological_inventory/schema/default.rb
```diff
@@ -123,6 +124,106 @@ def add_tagging_collection(model, manager_ref: [:source_ref])
...
+            service_instance_tasks_custom_save(source, tasks_collection)
+          end
+
+          builder.add_properties(:custom_save_block => save_block)
+        end
+      end
+
+      def service_instance_tasks_custom_save(source, tasks_collection)
+        service_instance_collection = tasks_collection.dependency_attributes[:service_instances]&.first
+        src_refs = service_instance_collection&.data.to_a.collect { |inventory_object| inventory_object.source_ref }
```
agrare, line 146: Should be able to do `service_instance_collection&.data&.collect(&:source_ref)` here

agrare:
I think the method names (effective/ineffective) are a little childish but I'll look past that so we can get this in, also should be changed in a follow-up.  I understand that you disagree with this approach but lets just try to move on and improve the project as a whole.

slemrmartin:
@agrare thanks for merging, I was quite unpleasant there, sorry about that

agrare:
Its alright we both have some strong opinions about this stuff so it is bound to happen, better than no opinions at all :smile:

PR #193, opened 2020-04-27
Saving updated tasks to IC

slemrmartin:
**Issue** https://github.com/RedHatInsights/topological_inventory-ansible_tower/issues/83

Updated tasks are saved to inventory collections for further processing by Persister's Workflow.

---

**dependent** https://github.com/RedHatInsights/topological_inventory-persister/pull/56

---

**Gem version: 1.1.2**
- [ ] released to **RubyGems**

### DIFF
--- a/db/migrate/20200428090420_add_request_id_to_task.rb
+++ b/db/migrate/20200428090420_add_request_id_to_task.rb
@@ -1,0 +1,5 @@
+class AddRequestIdToTask < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tasks, :x_rh_insights_request, :string
+  end
+end
--- a/lib/topological_inventory/schema/default.rb
+++ b/lib/topological_inventory/schema/default.rb
@@ -148,10 +148,10 @@ module TopologicalInventory
 
         # Updating Tasks
         # service_instance_tasks_update_by_raw_sql(source)
-        service_instance_tasks_update_by_activerecord(source, src_refs)
+        service_instance_tasks_update_by_activerecord(tasks_collection, source, src_refs)
       end
 
-      def task_update_values(svc_instance_id, external_url, status, task_status, finished_timestamp)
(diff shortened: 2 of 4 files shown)


db/migrate/20200428090420_add_request_id_to_task.rb
```diff
@@ -0,0 +1,5 @@
+class AddRequestIdToTask < ActiveRecord::Migration[5.2]
+  def change
+    add_column :tasks, :x_rh_insights_request, :string, :limit => 32
```
slemrmartin, line 3: I got the limit by observing dev Kibana. 
It would be good if somebody can ack if it is enough

agrare: Do we need to enforce a limit?

slemrmartin: right, this is varchar, not char. I'm still thinking in mysql :)

lib/topological_inventory/schema/default.rb
```diff
@@ -148,35 +148,42 @@ def service_instance_tasks_custom_save(source, tasks_collection)
...
 
-              values = task_update_values(svc_instance.id, svc_instance.external_url, svc_instance.extra['status'], svc_instance.extra['task_status'], svc_instance.extra['finished'])
-              tasks_by_source_ref[svc_instance.source_ref].update(values)
+              values = task_update_values(svc_instance.id, svc_instance.source_ref, svc_instance.external_url, svc_instance.extra['status'], svc_instance.extra['task_status'], svc_instance.extra['finished'], source.id)
+              # 1) Updating Task
+              task.update(values)
+
+              # 2) Saving to updated records (will be published in Kafka)
+              # - see topological_inventory-persister:Workflow.send_task_updates_to_queue!
+              tasks_collection.updated_records << values.merge(:id => task.id, :request_id => task.x_rh_insights_request)
```
agrare, line 186: > `request_id => task.x_rh_insights_request`

Should the key that we're setting here match the column name?

PR #204, opened 2020-08-11
Add Index to refresh_state_parts.uuid and github workflow for auto-releases

lindgrenj6:


db/migrate/20200811154716_add_index_to_refresh_state_parts.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndexToRefreshStateParts < ActiveRecord::Migration[5.2]
+  def change
+    add_index :refresh_state_parts, :uuid
```
slemrmartin, line 3: rather add also :refresh_state, :uuid

lindgrenj6: Added

lindgrenj6:
@syncrou this is ready for merge when green, the rubygems API key has been added by @gmcculloug

PR #205, opened 2020-09-10
Adding the 'acts_as_tenant' for RefreshState and RefreshStatePart models

pkomanek:
Adding the `acts_as_tenant` behaviour for RefreshState and RefreshStatePart models, since both of them inherit from the active_record and not from the application_record like rest of the models.

### DIFF
--- a/app/models/refresh_state.rb
+++ b/app/models/refresh_state.rb
@@ -1,6 +1,6 @@
-require 'acts_as_tenant'
-
 class RefreshState < ActiveRecord::Base
+  require 'acts_as_tenant'
+
   belongs_to :source
   belongs_to :tenant
   has_many :refresh_state_parts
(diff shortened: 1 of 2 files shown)


app/models/refresh_state.rb
```diff
@@ -1,3 +1,5 @@
+require 'acts_as_tenant'
```
slemrmartin, line 1: @pkomanek should be rather included inside of a class like `ApplicationRecord` has

app/models/refresh_state_part.rb
```diff
@@ -1,3 +1,5 @@
+require 'acts_as_tenant'
```
slemrmartin, line 1: same as above

slemrmartin (changes requested):

slemrmartin (approved):

PR #206, opened 2020-09-10
Changelog and gem version upgrade.

pkomanek:
Upgrading the gem version and CHANGELOG file.

- [ ] Depends on #205


slemrmartin (approved):
It seems there was missing changelog change for 1.1.6.

PR #207, opened 2020-09-16
Added artifacts to tasks IC and bulk sql update

slemrmartin:
Exposing ServiceInstance.extra['artifacts'] to Task.context['artifacts'].
Also switched from ActiveRecord update to bulk SQL update (50 times faster on 1000 records)

---

[TPINVTRY-1079](https://issues.redhat.com/browse/TPINVTRY-1079)

### DIFF
--- a/lib/topological_inventory/schema/default.rb
+++ b/lib/topological_inventory/schema/default.rb
@@ -147,8 +147,8 @@ module TopologicalInventory
         return if src_refs.blank?
 
         # Updating Tasks
-        service_instance_tasks_update_by_raw_sql(source)
-        # service_instance_tasks_update_by_activerecord(tasks_collection, source, src_refs)
+        # service_instance_tasks_update_by_raw_sql(source)
+        service_instance_tasks_update_by_activerecord(tasks_collection, source, src_refs)
       end
 
       def task_update_values(svc_instance_id, source_ref, external_url, status, artifacts, task_status, finished_timestamp, source_id)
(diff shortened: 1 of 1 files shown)


lib/topological_inventory/schema/default.rb
```diff
@@ -147,18 +147,19 @@ def service_instance_tasks_custom_save(source, tasks_collection)
         return if src_refs.blank?
 
         # Updating Tasks
-        # service_instance_tasks_update_by_raw_sql(source)
-        service_instance_tasks_update_by_activerecord(tasks_collection, source, src_refs)
+        service_instance_tasks_update_by_raw_sql(source)
```
syncrou, line 150: I think we can pull this change out of the PR.  I understand the reasoning for wanting to put it in, but it falls outside the changes specific to exposing the artifacts on a service_instance.

I  would be interested in moving to the raw sql method when there is a clear performance bottleneck we need to resolve.

syncrou (changes requested):
This looks good, if we can pull out the sql caller change I think it will be good to go.

lib/topological_inventory/schema/default.rb
slemrmartin: @syncrou whole refresh is built on raw sql, this is one of the few exceptions - https://github.com/ManageIQ/inventory_refresh/blob/master/lib/inventory_refresh/save_collection/saver/sql_helper_upsert.rb
Why do we need to wait for bottleneck and then solve it (and spend many hours with diagnostics probably on production)? Isn't it better to prevent it?
Ok, I'll remove it from this PR, but are you opened to move it to new PR?

syncrou (approved):
Thanks @slemrmartin! LGTM 👍

PR #211, opened 2020-10-27
Rubocop rules from insights-api-common

slemrmartin:
Because of https://github.com/ManageIQ/guides/pull/443 rubocop rules were moved to another repo. 
So we are moving the file to our common repo.

* [x] depends on https://github.com/RedHatInsights/insights-api-common-rails/pull/210

---

[RHCLOUD-10237](https://issues.redhat.com/browse/RHCLOUD-10237)

PR #212, opened 2020-10-29
Rubocop rules from insights-api-common + yamllint

slemrmartin:
Because of https://github.com/ManageIQ/guides/pull/443 rubocop rules were moved to another repo. 
So we are moving the file to our common repo.

* extending #211
* [x] depends on https://github.com/RedHatInsights/insights-api-common-rails/pull/211

---

[RHCLOUD-10237](https://issues.redhat.com/browse/RHCLOUD-10237)

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -2,9 +2,9 @@
 version: '2'
 prepare:
   fetch:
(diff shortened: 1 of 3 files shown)


lindgrenj6 (approved):
Good to know we can just add dev dependencies on rubocop 1.0/performance/rails. Just wish we could bulk update all the repos!

PR #217, opened 2021-01-19
Added missing index RefreshState:updated_at for DB cleaner

slemrmartin:
Topological-inventory-db-cleaner is using:
`RefreshState.where('updated_at <= ?', threshold).delete_all`

but the db index is missing so it takes ages 

I thought we solved this issue before...


syncrou (approved):
LGTM 👍

lindgrenj6 (commented):
So I'm not sure we'll need this after the initial HUGE deletion. 

I think we should let the job run for a long time, see how the job runs on the second time through. If it's still pretty slow then we'll add the index.

slemrmartin:
@lindgrenj6 the Job haven't finished (it's running for 17 hours)

lindgrenj6 (approved):

PR #218, opened 2021-03-10
Define Task model in migration ExtractTasksSourceRefFromContext

lpichler:
This change keeps migrations independent  on `app/models`.


db/migrate/20200414123737_extract_tasks_source_ref_from_context.rb
```diff
@@ -1,4 +1,6 @@
 class ExtractTasksSourceRefFromContext < ActiveRecord::Migration[5.2]
+  class Task < ApplicationRecord; end
```
slemrmartin: I guess it is needed only when migrations and models are in different repos

lpichler: It is needed also in this repo as we can add various callbacks/validation or other surprises later, which could affect data migrations unintentionally. This is reason why it is better to have defined models here in minimal version as much as possible.

